PR #5, opened 2020-04-20
Add metal-api gRPC certificates for new machine allocation

kolsa:
References https://github.com/metal-stack/metal-api/pull/55.


control-plane/roles/nsq/templates/nsq.yaml
```diff
@@ -73,7 +73,7 @@ spec:
         args:
           - /nsqadmin
           - -log-level={{ nsq_log_level }}
-          - -lookupd-http-address=nsq-lookupd:4161
+          - -lookupd-http-address={{ api_nsq_lookupd }}
```
Gerrit91, line 76: This uses a variable of another role, which violates role separation. 

What was the reason for you to change it? The nsq lookupd service has a static name as well. Can this even work with another http address without special DNS records?

partition/roles/grpc-proxy/files/grpc-proxy.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=gRPC Proxy
+After=docker.service
+Wants=docker.service
+StartLimitIntervalSec=0
+
+[Service]
+ExecStartPre=-/usr/bin/docker pull nginx:1.17
```
Gerrit91, line 8: Is there a specific reason why you did not use the `ansible-common/roles/systemd-docker-service` as done by the other services? This would make all systemd service files equivalent across the deployment.

partition/roles/metal-core/defaults/main.yaml
```diff
@@ -11,6 +11,14 @@ metal_core_nsq_cert_dir: /certs/nsq
 metal_core_nsqd_ca_cert:
 metal_core_nsqd_client_cert:
 
+metal_core_grpc_port:
```
Gerrit91, line 14: You can possibly default this, right?

PR #6, opened 2020-04-23
Properly maintain role separation for service exposal by merging CMs.

Gerrit91:
We did not find a way to expose services from the individual roles in the past and put everything into the metal role. It's not that hard to do this from individual roles though by using the `combine` filter from ansible.


mwindower:
lgtm

PR #9, opened 2020-05-06
changed docker-on-cumulus role: docker install for cumulus release 3&4

Donimax:
- added major version verification 
- added template for docker systemd service

### DIFF
--- a/partition/roles/docker-on-cumulus/tasks/main.yaml
+++ b/partition/roles/docker-on-cumulus/tasks/main.yaml
@@ -1,66 +1,8 @@
 ---
-- name: setup key for docker
-  apt_key:
-    url: https://download.docker.com/linux/ubuntu/gpg
-    id: 9DC858229FC7DD38854AE2D88D81803C0EBFCD88
+- name: install docker on cumulus buster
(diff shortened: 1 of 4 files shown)


partition/roles/docker-on-cumulus/tasks/main.yaml
```diff
@@ -1,66 +1,8 @@
 ---
-- name: setup key for docker
-  apt_key:
-    url: https://download.docker.com/linux/ubuntu/gpg
-    id: 9DC858229FC7DD38854AE2D88D81803C0EBFCD88
+- name: install docker on cumulus buster
+  import_tasks: cumulus_buster.yaml
+  when: ansible_facts['distribution_major_version'] == "4"
```
Gerrit91, line 4: What about the injected `ansible_distribution_major_version` without dictionary lookup? Wouldn't this throw a nicer error message?

Donimax: Both vars `ansible_facts['distribution_major_version']` and `ansible_distribution_major_version`  are working. yes it's little bit easier to read if the variable is undefined ;)

PR #13, opened 2020-06-04
Further refactoring

Gerrit91:
Fixes some bugs along the way.

- Variable renaming
  - strictly prefixing roles and partition defaults
- Now allows deployments without NSQ TLS encryption as well
- Documents all variables from the partition defaults -> now prefixed with `metal_partition_`
- Adds metal-core role to partition documentation
- Refactoring of dhcp role
  - added documentation
  - now also ensuring that service is started after deployment
  - not deployed "when" a certain variable gets defined, needs to be targeted explicitly from now on
- Disabling fluentd Docker daemon logging by default, no explicit configuration required anymore
- Reconfigure switch is now enabled by default

Not contained: 
- Documentation of how the partition inventory needs to look like


Gerrit91:
Can be tried out from the following mini-lab PR: https://github.com/metal-stack/mini-lab/pull/15

PR #23, opened 2020-08-20
bmc-proxy-endpoints are now external name services.

Gerrit91:
Load balancing must be provided at the destination, not in our control plane cluster.


kolsa (approved):
lgtm

PR #40, opened 2020-11-11
configure ipmi superuser credentials for bmc catcher

kolsa:


kolsa:
Closes https://github.com/metal-stack/bmc-catcher/issues/13

partition/roles/bmc-catcher/tasks/main.yaml
```diff
@@ -29,3 +29,5 @@
       BMC_CATCHER_METAL_API_URL: "{{ metal_partition_metal_api_protocol }}://{{ metal_partition_metal_api_addr }}:{{ metal_partition_metal_api_port }}{{ metal_partition_metal_api_basepath }}"
       BMC_CATCHER_METAL_API_HMAC_KEY: "{{ metal_partition_metal_api_hmac_edit_key }}"
       BMC_CATCHER_IGNORE_MACS: "{{ bmc_catcher_ignore_macs | join(',') }}"
+      BMC_CATCHER_IPMI_USER: root
+      BMC_CATCHER_IPMI_PASSWORD: "{{ metal_api_bmc_superuser_pwd }}"
```
Gerrit91, line 33: Mh... I made the constraint for roles not to use variables from other roles in order to make them fully independent. If you want to re-use this variable, you could add them to the global defaults and call them `metal_stack_bmc_superuser_pwd`. Or you repeat the variable, in which case it would be called `bmc_catcher_bmc_superuser_pwd`. I would also make the user variable, I think.

Gerrit91 (changes requested):

Gerrit91 (approved):

PR #60, opened 2021-04-28
MEP-8

majst01:


control-plane/roles/metal/README.md
```diff
@@ -89,6 +89,7 @@ You can look up all the default values of this role [here](defaults/main/main.ya
 | metal_api_partitions             |           | Creates partitions (as masterdata) to the metal-api after deployment |
 | metal_api_networks               |           | Creates networks (as masterdata) to the metal-api after deployment   |
 | metal_api_ips                    |           | Creates ips (as masterdata) to the metal-api after deployment        |
+| metal_api_filesystemlayouts      | yes       | Creates filesystemlayouts  to the metal-api after deployment         |
```
Gerrit91, line 92: When this variable defaults to something (in this case empty list) this is not required for a user to set. Required variables are also checked at the start of the role such that deployment fails immediately when not provided.

```suggestion
| metal_api_filesystemlayouts      |           | Creates filesystemlayouts to the metal-api after deployment          |
```

Gerrit91 (approved):

PR #72, opened 2021-11-15
Adding metal-python installation role.

Gerrit91:


GrigoriyMikhalkin (approved):
lgtm.

I guess, after merging this we also need to review metal-python installation role from mini-lab.

PR #91, opened 2022-06-23
move bmc code from metal-core to bmc-catcher

majst01:
depends on:

- [ ] https://github.com/metal-stack/bmc-catcher/pull/33
- [ ] https://github.com/metal-stack/metal-core/pull/72


partition/roles/bmc-catcher/defaults/main/main.yaml
```diff
@@ -3,3 +3,10 @@ bmc_catcher_ignore_macs: []
 
 bmc_catcher_bmc_superuser:
 bmc_catcher_bmc_superuser_pwd:
+
+
+bmc_catcher_nsq_log_level: warn
```
Gerrit91, line 8:
```suggestion
bmc_catcher_nsq_log_level: info
```

partition/roles/bmc-catcher/tasks/main.yaml
```diff
@@ -12,12 +12,35 @@
       - bmc_catcher_bmc_superuser is defined
       - bmc_catcher_bmc_superuser_pwd is defined
```
Gerrit91, line 14: Add two assertions: 
```suggestion
      - not bmc_catcher_nsq_tls_enabled or (bmc_catcher_nsq_tls_enabled and bmc_catcher_nsqd_ca_cert is not none)
      - not bmc_catcher_nsq_tls_enabled or (bmc_catcher_nsq_tls_enabled and bmc_catcher_nsqd_client_cert is not none)
```

Gerrit91 (changes requested):
NSQ certs still have to be mounted into the docker container.

Gerrit91 (approved):

partition/roles/metal-core/defaults/main/main.yaml
```diff
@@ -4,12 +4,6 @@ metal_core_hmac_key: change-me
 metal_core_rack_id:
 metal_core_change_boot_order: true
```
Gerrit91, line 5: This variable was removed and should also be removed here

PR #103, opened 2022-09-15
increment headscale retries

GrigoriyMikhalkin:
Incremented Headscale start retries, otherwise mini-lab often fails with time outs.


Gerrit91 (approved):

Gerrit91:
Do you wan to make a release of this or will there be more changes to come such that we can bundle them together?

GrigoriyMikhalkin:
> Do you wan to make a release of this or will there be more changes to come such that we can bundle them together?

I think we can make a release

Gerrit91:
Done.

PR #109, opened 2022-11-18
Make headscale private key conditional.

Gerrit91:


Gerrit91:
The private key is required by the headscale server, cannot be removed.

PR #112, opened 2022-12-16
Providing roles for partition management network bootstrapping

Gerrit91:


control-plane/roles/metal-python/defaults/main.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-metal_python_version:
-metal_python_version_from_release_vector: yes
+metal_python_version: v0.21.4
+metal_python_version_from_release_vector: no
```
Gerrit91, line 3: Needs to be reverted

partition/roles/dhcp/tasks/main.yaml
```diff
@@ -4,26 +4,60 @@
...
-      - dhcp_range_min is not none
-      - dhcp_range_max is not none
+      - item.network is defined
+      - item.netmask is not none
+      - item.range is not none
+      - item.range.begin is not none
+      - item.range.end is not none
+  loop: "{{ dhcp_subnets }}"
+
+- name: Get stat for SONiC's version file
```
Gerrit91, line 14: The role should not be aware of operating systems. Looks like the only purpose is to add the `force_apt_get: yes` parameter? Why not just pass it explicitly as a role parameter?

Gerrit91:
This has become way too big. I will close this and open smaller PRs in order to be able to review and merge.

PR #87, opened 2022-06-17
SONiC

robertvolkmann:

PR #121, opened 2023-01-23
Add `sonic` role to support SONiC switches.

Gerrit91:
I gathered the relevant changes for SONiC support in this new PR.

The work on this PR started in https://github.com/metal-stack/metal-roles/pull/112 and https://github.com/metal-stack/metal-roles/pull/87. It supersedes both PRs.


partition/roles/sonic/tasks/main.yaml
```diff
@@ -0,0 +1,65 @@
...
+  assert:
+    fail_msg: "not all mandatory variables given, check role documentation"
+    quiet: yes
+    that:
+      - sonic_mgmt_vrf is defined
+      - sonic_loopback_address is defined
+      - sonic_asn is defined
+      - sonic_ntpservers is defined
+      - sonic_nameservers is defined
+
```
Gerrit91, line 12: It's also useful to run `switch_facts` and fail if the switch os is not Sonic.

partition/roles/dhcp/templates/dhcpd.conf.j2
```diff
@@ -15,13 +15,13 @@ subnet {{ subnet.network }} netmask {{ subnet.netmask }} {
   range {{ subnet.range.begin }} {{ subnet.range.end }};
 {% endif %}
 {% for option in subnet.options | default([]) %}
-  option {{ option }};
```
Gerrit91, line 18: The interface should not be changed again, this would be breaking for mini-lab. Just omit the `option` string when defining the options.

Gerrit91: Also option is actually required as otherwise a user can mess up the config with arbitrary values easily.

Gerrit91: I reverted these changes.

partition/roles/metal-core/files/metal-core.timer
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Delays startup of metal-core until SONiC has started properly
+
+[Timer]
+OnUnitActiveSec=0 sec
+OnBootSec=3min 30 sec
```
robertvolkmann, line 6: `metal-core` should handle it on its own instead of introducing some random interval.

partition/roles/mgmt-server/tasks/main.yaml
```diff
@@ -85,9 +87,66 @@
...
+  - content: "{{ mgmt_server_metal_ssh_privkey }}"
+    dest: /home/metal/.ssh/id_rsa
+    mode: "0600"
+  - content: "{{ mgmt_server_metal_ssh_pubkey }}"
+    dest: /home/metal/.ssh/id_rsa.pub
+    mode: "0644"
+
+# This is so that self connect and cross connect to the other mgmtserver is possible
+- name: Add own ssh key to authorized_keys
+  lineinfile:
```
robertvolkmann, line 107: Why not using `ansible.posix.authorized_key`?

partition/roles/mgmt-server/tasks/main.yaml
```diff
@@ -85,9 +87,66 @@
...
+    mode: "{{ item.mode}}"
+    owner: metal
+    group: metal
+  loop:
+  - content: "{{ mgmt_server_metal_ssh_privkey }}"
+    dest: /home/metal/.ssh/id_rsa
+    mode: "0600"
+  - content: "{{ mgmt_server_metal_ssh_pubkey }}"
+    dest: /home/metal/.ssh/id_rsa.pub
+    mode: "0644"
```
robertvolkmann, line 103: Why we need the public key on the machine?

partition/roles/mgmt-server/tasks/main.yaml
```diff
@@ -85,9 +87,66 @@
...
+  lineinfile:
+    dest: /etc/hosts
+    regexp: '.*\s{{ item }}$'
+    line: "{{ hostvars[item].ansible_host }} {{ item }} {{ hostvars[item].host_alias|default('') }}"
+    owner: root
+    group: root
+    mode: 0644
+  when:
+    - hostvars[item].ansible_host is defined
+  with_items: "{{ groups.all }}"
```
robertvolkmann, line 135: Should be injected as variable instead of using all `groups.all`. Is an unexpected behavior of the role.

partition/roles/frr/tasks/install/debian.yaml
```diff
@@ -0,0 +1,19 @@
+---
+- name: add FRR apt-key to verify FRR package
+  apt_key:
+    url: https://deb-us.frrouting.org/frr/keys.asc
+    state: present
+
+- name: add mainline FRR repository to install FRR
+  apt_repository:
+    repo: deb https://deb-us.frrouting.org/frr {{ ansible_distribution_release }} {{ mgmt_server_frr_repo }}
```
Gerrit91, line 9: There should be no cross-usage between ansible role variables.

partition/roles/mgmt-server/templates/ssh_config.j2
```diff
@@ -0,0 +1,8 @@
+{% for host in groups.all %}
```
Gerrit91, line 1: This loop can be very exhaustive. Instead provide a `mgmt_server_ssh_config_groups: []` variable and let the user decide which hosts to add.

partition/roles/mgmt-server/tasks/install_frr.yaml
```diff
@@ -1,4 +1,11 @@
 ---
+- name: ensure gpg is installed to handle FRR repo key
```
Gerrit91, line 2: It's a bit sad we have an FRR role now and the mgmt-server's FRR is still installed here without using it.

partition/roles/mgmt-server/tasks/main.yaml
Gerrit91: I don't know.

partition/roles/sonic/tasks/main.yaml
```diff
@@ -0,0 +1,99 @@
...
+  with_items:
+    - bgp-validation@.service
+    - frr-reload.service
+
+# Dependecies is returned by config.
+- name: Configure breakouts
+  command: "config interface breakout --yes {{ item.key }} '{{ item.value }}'"
+  register: breakout_result
+  changed_when: "'Breakout process got successfully completed.' in breakout_result.stdout"
+  failed_when: "breakout_result.rc != 0 or 'Dependecies Exist. No further action will be taken' in breakout_result.stdout"
```
Gerrit91, line 66: `Dependecies Exist`? Typo in SONiC?

partition/roles/systemd-networkd/tasks/main.yaml
```diff
@@ -29,7 +38,7 @@
   when: item.name != 'lo'
   notify:
   - restart systemd-networkd
-  - restart systemd-udev-trigger
+  - rename interfaces
```
Gerrit91, line 41: This handler does not exist?

partition/roles/sonic/tasks/main.yaml
robertvolkmann: Yeah, a typo in SONiC.

partition/roles/systemd-networkd/tasks/main.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-- include_tasks: redhat.yaml
+- import_tasks: redhat.yaml
```
robertvolkmann, line 2: Why `import_` instead of `include_`?

partition/roles/systemd-networkd/tasks/main.yaml
mreiger: There's two handlers that listen for that event:

https://github.com/metal-stack/metal-roles/blob/11aeaf21135c0f0e5b3788f87721074012090b4f/partition/roles/systemd-networkd/handlers/main.yaml#L32
https://github.com/metal-stack/metal-roles/blob/11aeaf21135c0f0e5b3788f87721074012090b4f/partition/roles/systemd-networkd/handlers/main.yaml#L38

The point for this is:
The renaming of interfaces does not take effect while they're still up; so the following deployment steps that depend on the interface names would fail. 
With this procedure the interfaces are taken offline and the renaming gets triggered.

partition/roles/systemd-networkd/tasks/main.yaml
Gerrit91: Import enables Ansible to compile the task execution at the start, include adds them dynamically during the execution. If there is a syntax error in `redhat.yaml` for example, Ansible will fail immediately at the start when using `import` instead of dying in the middle of the run when the include happens, which I think is beneficial.

Here are some more tradoffs: https://docs.ansible.com/ansible/2.9/user_guide/playbooks_reuse.html#tradeoffs-and-pitfalls-between-includes-and-imports

partition/roles/systemd-networkd/tasks/main.yaml
Gerrit91: Thanks for explaining. I was not aware of the `listen` keyword.

vknabel (approved):

PR #139, opened 2023-02-07
Add wireguard role to partition

mwindower:


majst01:
Unsure if we should add this here, i would like to prefer the tailscale approach. But does not hurt.

majst01 (approved):

partition/roles/wireguard/tasks/main.yaml
```diff
@@ -0,0 +1,99 @@
+---
+- name: check mandatory variables for this role are set
+  assert:
+    fail_msg: "not all mandatory variables given, check role documentation"
+    quiet: yes
+    that:
+      - wireguard_ip is not none
+      - wireguard_clients is not none
+      - wireguard_clients|length > 0
```
Gerrit91, line 9: Just `wireguard_clients` is enough, it does exactly this

partition/roles/wireguard/tasks/main.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+- name: install wireguard
+  apt:
+    name: wireguard
+    update_cache: yes
+    state: latest
+  with_items:
```
Gerrit91, line 16: Deprecated for `loop`

partition/roles/wireguard/tasks/main.yaml
```diff
@@ -0,0 +1,99 @@
...
+    - wireguard
+    - wireguard-dkms
+    - wireguard-tools
```
Gerrit91, line 19: `item` is not used, this loop does three times the same?

partition/roles/wireguard/tasks/main.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+  - name: set public key fact
+    set_fact:
+      wg_public_key: "{{ wg_public_key_result.stdout }}"
+
+  when: not private_key_file_stat.stat.exists
+  tags: wg-generate-keys
+
+- name: write WireGuard configuration
+  block:
```
Gerrit91, line 75: `block` is used but no reason to do so

partition/roles/wireguard/tasks/main.yaml
```diff
@@ -0,0 +1,99 @@
...
+- name: install wireguard
+  apt:
+    name: wireguard
+    update_cache: yes
+    state: latest
+  with_items:
+    - wireguard
+    - wireguard-dkms
+    - wireguard-tools
+  tags: wg-install
```
Gerrit91, line 20: Do we need the tags on these tasks? Somehow we do not have any tags anywhere else in the roles of this repo.

PR #140, opened 2023-02-07
Add gitlabci-runner role to partition

mwindower:


majst01 (approved):

partition/roles/gitlabci-runner/tasks/main.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - gitlab_runner_url is not none
+      - gitlab_runner_token is not none
+      - gitlab_runner_description is not none
+      - gitlab_runner_tags is not none
+      - gitlab_runner_hostpath is not none
+
+- name: install docker as a prerequisite to gitlabci
+  apt:
+    name:
+    - docker.io
```
Gerrit91, line 16: Use a dedicated docker role to do this, the role name does not imply this dependency. There are a good roles in ansible-galaxy to do this.

Gerrit91 (commented):
Maybe it makes sense to use this role here instead: https://galaxy.ansible.com/riemers/gitlab-runner

PR #145, opened 2023-02-20
Auditing metal_auditing_namespace

vknabel:
Allows alternative namespaces for the auditing.

### DIFF
--- a/control-plane/roles/auditing-meili/tasks/main.yaml
+++ b/control-plane/roles/auditing-meili/tasks/main.yaml
@@ -8,9 +8,9 @@
       apiVersion: v1
       kind: Namespace
       metadata:
-        name: "{{ metal_auditing_namespace }}"
+        name: "{{ auditing_meili_namespace }}"
         labels:
-          name: "{{ metal_auditing_namespace }}"
+          name: "{{ auditing_meili_namespace }}"
 
 - name: Create meili secret
   k8s:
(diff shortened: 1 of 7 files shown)


control-plane/roles/auditing-meili/tasks/main.yaml
```diff
@@ -19,18 +19,30 @@
...
+  k8s:
+    definition:
+      apiVersion: v1
+      kind: Secret
+      metadata:
+        name: metal-auditing-master-key
+        namespace: "{{ metal_control_plane_namespace }}"
+      stringData:
+        MEILI_MASTER_KEY: "{{ auditing_meili_secret }}"
+  when: auditing_meili_secret and (metal_control_plane_namespace != metal_auditing_namespace)
```
Gerrit91, line 37: This appears a little too specific to me / it is introducing some kind of a role coupling.

My suggestion would be not to use `metal_auditing_meili_secret_name` in the `metal` role. Instead use `metal_auditing_meili_api_key` and do this in the defaults:

```
metal_auditing_meili_api_key: "{{ lookup('k8s', api_version='v1', namespace=metal_auditing_namespace, kind='Secret', resource_name='metal-auditing-master-key').get('data', {}).get('MEILI_MASTER_KEY') | b64decode if metal_auditing_enabled else '' }}"
```

This is already done by the headscale role, too. This will make your role a bit more self-contained. :)

vknabel: Thanks for the suggestion!

Gerrit91 (approved):

PR #147, opened 2023-02-22
Rename config save reload systemd service

GrigoriyMikhalkin:
Also, removes `config save -y` from pre-start. It will be put in another place later.


partition/roles/sonic/tasks/main.yaml
```diff
@@ -47,7 +47,7 @@
   with_items:
     - bgp-validation@.service
     - write-to-db@.service
-    - config-save-reload.service
+    - frr-reload.service
```
majst01, line 50: This needs adoptions in metal-core as well

majst01 (approved):

PR #163, opened 2023-04-14
Update sonic params for metal-core.

Gerrit91:


partition/roles/metal-core/tasks/main.yaml
```diff
@@ -52,6 +52,7 @@
     systemd_docker_volumes: "{{ lookup('template', 'metal-core-volumes.j2') | from_yaml }}"
     systemd_docker_cap_add:
       - sys_admin
+      - net_admin
```
robertvolkmann, line 55:
```suggestion
```
Not needed anymore.

partition/roles/metal-core/templates/metal-core-volumes.j2
```diff
@@ -1,6 +1,6 @@
 {% if metal_stack_switch_os_is_sonic %}
 - /etc/sonic/:/etc/sonic
-- /var/run/redis/:/var/run/redis:ro
+- /var/run/redis/:/var/run/redis
```
robertvolkmann, line 3: I don't understand the reason of this change.

robertvolkmann (changes requested):

partition/roles/metal-core/templates/metal-core-volumes.j2
Gerrit91: When writing into the database it cannot be r/o anymore?

partition/roles/metal-core/tasks/main.yaml
Gerrit91: Perfect

partition/roles/metal-core/templates/metal-core-volumes.j2
robertvolkmann: Good point.

robertvolkmann (approved):

robertvolkmann:
What about the `README.md`?

partition/roles/metal-core/tasks/main.yaml
```diff
@@ -14,6 +14,10 @@
       - metal_core_image_tag is defined
       - metal_core_rack_id is not none
 
+- name: Set service dependency
+  set_fact:
+    metal_core_service_dependency: "{{ database.service if metal_stack_switch_os_is_sonic else docker.service }}"
```
Gerrit91, line 19: I think this has to be quoted, otherwise in the Jinja context it will be interpreted as a variable:

```suggestion
    metal_core_service_dependency: "{{ 'database.service' if metal_stack_switch_os_is_sonic else 'docker.service' }}"
```

PR #165, opened 2023-04-19
feat: meili v1.1.1

vknabel:
Requires https://github.com/metal-stack/releases/pull/114


majst01:
Not sure if we should put another entry into the release vector: https://github.com/metal-stack/releases/blob/master/release.yaml#L189

vknabel:
Yes, we should. I'll add it. Somehow I thought the release mapping would be in the mini-lab. I'll add it.

majst01 (approved):

Gerrit91:
When the proper container image versions are already shipped with their chart releases, why do we manage them individually now? I can understand overriding these optionally but the default version should come from their chart releases.

vknabel:
The Multisearch we use to list auditing traces is a v1.1 feature, but the latest chart release still defaults to v1.0

PR #179, opened 2023-06-15
Fix regression for gardener dns-external extension.

Gerrit91:


majst01:
Isnt that fixed with a newer version of the extension ?

Gerrit91:
I am not sure. It's happening because our seed clusters are too old and I don't think they are aware of this regression. The extension should be removed in the future anyway.

majst01 (approved):

PR #193, opened 2023-07-14
fix: fsl type parsing

vknabel:
Different YAML parsing implementations handle literals like `8e00` differently. Enforcing quotes with jsonsolves this issue.


Gerrit91:
You can do this for all entities, I think it's safer?

vknabel:
Now JSON is used for all entities in this file

Gerrit91 (approved):

PR #195, opened 2023-07-17
Add possibility to deploy Gardener ingress certs.

Gerrit91:
Should resolve no certs on shoot Grafana dashboards.


majst01 (approved):

control-plane/roles/gardener-monitoring-certs/tasks/main.yaml
```diff
@@ -0,0 +1,51 @@
...
+- name: Add ingress certificate for shooted seeds
+  k8s:
+    definition:
+      apiVersion: cert-manager.io/v1
+      kind: Certificate
+      metadata:
+        name: seed-ingress
+        namespace: garden
+        labels:
+          use-clouddns-solver: "true"
```
robertvolkmann, line 11: I don't understand this label.

Gerrit91: Probably it's too opinionated because it assumes Google CloudDNS to be used. In general, it can be used for the `matchLabels` selector in the `ClusterIssuer` solver.

PR #196, opened 2023-07-19
Fix default pull secret registry

Gerrit91:
This seems to fix our CI not being able to use our Docker Hub user. With this PR this seems fixed: https://github.com/metal-stack/mini-lab/actions/runs/5600457774/jobs/10242992559

(this runner reached the pull limits previously here: https://github.com/metal-stack/releases/actions/runs/5600206508/jobs/10242259193)


Gerrit91:
I am not 100% sure if this works, but it does not work anyway, so it's an okay-ish attempt.

Gerrit91:
Not seen anymore, re-open if necessary.

PR #198, opened 2023-07-25
Changes for g/g v1.58 compatibility.

Gerrit91:
```ACTIONS_REQUIRED
Seeds were deployed with an incorrect `PodCIDR` in the past, which in Gardener 1.58 leads to a validation failure for `Seed` resources. Prior to updating to this version, the pod CIDR needs to be patched to the correct CIDR. This can be achieved through a modified gardener-apiserver image (`r.metal-stack.io/gardener/gardener-apiserver:remove-pod-cidr-validation`), built from [this fork branch](https://github.com/gardener/gardener/compare/v1.57.2...metal-stack:gardener:remove-pod-cidr-validation).
```

### DIFF
--- a/control-plane/roles/gardener/tasks/control-plane-networks/gcp.yaml
+++ b/control-plane/roles/gardener/tasks/control-plane-networks/gcp.yaml
@@ -34,7 +34,7 @@
 
 - name: Set pod cidr
   set_fact:
-    _gardener_soil_pod_cidr: "{{ lookup('k8s', kind='Node')[0].get('spec', {}).get('podCIDR') }}"
+    _gardener_soil_pod_cidr: "{{ (gcp_cluster.stdout | from_json)['ipAllocationPolicy']['clusterIpv4CidrBlock'] }}"
 
 - name: Set service cidr
   set_fact:


control-plane/roles/gardener/tasks/control-plane-networks/gcp.yaml
```diff
@@ -34,7 +34,7 @@
 
 - name: Set pod cidr
   set_fact:
-    _gardener_soil_pod_cidr: "{{ lookup('k8s', kind='Node')[0].get('spec', {}).get('podCIDR') }}"
+    _gardener_soil_pod_cidr: "{{ (gcp_cluster.stdout | from_json)['ipAllocationPolicy']['clusterIpv4CidrBlock'] }}"
```
majst01, line 37: the chances below are also required

majst01 (changes requested):

control-plane/roles/gardener/tasks/control-plane-networks/gcp.yaml
Gerrit91: This deployment role is a bit different from the one used by FITS, so this is not required here.

PR #201, opened 2023-08-20
Support Postgres Upgrades

majst01:


majst01:
This is obsolete now, the backup-restore-sidecar manages this without any modifications in the manifest

PR #232, opened 2023-11-08
Ensure AgentX is enabled on leaves

robertvolkmann:
Newer versions of Edgecore SONiC enable AgentX protocol support by default. It is hardcoded in the supervisor configuration in the bgp container and once enabled, it cannot be unconfigured anymore. As result, we activate it on every SONiC distribution.


robertvolkmann:
Should be handled by #358.

PR #245, opened 2023-12-12
Refactor Services Dependency from 'sonic' to 'metal-core' for Broadcom Compatibility

robertvolkmann:
This pull request revises the service dependencies of `metal-core`, detaching it from the `sonic` role and integrating necessary services directly into `metal-core`. This update addresses the hard dependency issue that previously existed, and it extends the compatibility of `metal-core` to not just Edgecore but now also the Broadcom distribution of SONiC.


partition/roles/sonic/handlers/main.yaml
```diff
@@ -1,8 +1,4 @@
 ---
-- name: reload systemd
```
majst01, line 2: this name is nowhere else used ?

robertvolkmann: Is unused in the role `sonic` after moving affected parts to `metal-core`. In the role `metal-core`, I renamed the handler to `Reload systemd daemon`.

majst01 (approved):

PR #254, opened 2024-01-17
Introduce metal-core sync schedule.

Gerrit91:


Gerrit91:
Decided that it does not bring the expected benefit, so we can close this.

Issue #235, opened 2023-11-20
Create roles for dns, ntp and registry


majst01:
Move existing roles from metal-deployment into dedicated roles here

Related to: https://github.com/fi-ts/proxy-services/issues/4

Requirements for the registry:

- Deployed on every seed
- external access restricted by the usage of loadbalancer source ranges
- Make the image overrides configurable
- Use persistent volumes as backing store
- Multiple instances with same ip, use ECMP for loadbalancing

DNS:
- ...

NTP:
- ...

Gerrit91:
This does not need to go into v0.17.0 release for isolated cluster support but for the community it would be great to push this automation into this repository.

Gerrit91:
Done.

PR #259, opened 2024-02-06
feat(gardener): network isolation #235

vknabel:
Passes through network isolation if given for partition #235

Do I miss some changes?


majst01:
> Passes through network isolation if given for partition #235
> 
> Do I miss some changes?

Yes the whole deployment of the oci-mirror and the registry, the dns and ntp server with the ingress spreading the load across

control-plane/roles/gardener/defaults/main/cloud_profile.yaml
```diff
@@ -14,6 +14,23 @@ gardener_cloud_profile_partitions: {}
   #   default-machine-types:
   #     firewall:
   #       - c1-xlarge-x86
+  #     network-isolation:
```
mwennrich, line 17: should not be within default-machine-types

control-plane/roles/isolated-clusters/tasks/dns.yaml
```diff
@@ -0,0 +1,27 @@
+- name: Get kubeconfig
+  copy:
+    src: "files/kubeconfigs/kubeconfig.{{ cluster }}"
```
Gerrit91, line 3: About this convention I am not so sure if I am happy. This would at least need to be documented, but then it also motivates long-lived kubeconfigs to be statically put into the deployment, which is probably not so ideal.

As the entire cluster isolation feature requires Gardener anyway, maybe we could do it similar to the way roles like [gardener-monitoring-certs](https://github.com/metal-stack/metal-roles/tree/master/control-plane/roles/gardener-monitoring-certs) do it. So the virtual garden kubeconfig is looked up from to control plane cluster (for which the kubeconfig is usually dynamically created). 

From the virtual garden you can dynamically create an admin kubeconfig for a specific shoot (so `isolated_clusters_partition_services_cluster` needs to be a list of dicts with `shoot_name` and `project_name`). This would work with this filter plugin (the way how to do it is actually taken from gardener docs [here](https://github.com/gardener/gardener/blob/297c6c19ab33751bb27c28ab3d1e41d3706ea918/docs/usage/shoot_access.md#shootsadminkubeconfig-subresource)):

```python
from __future__ import (absolute_import, division, print_function)

__metaclass__ = type

import base64
import yaml
import json

from kubernetes import config, client


def shoot_admin_kubeconfig(kubeconfig, project_namespace, shoot_name, expiry_seconds=28800):
    data = yaml.safe_load(kubeconfig)

    configuration = client.Configuration()
    config.load_kube_config_from_dict(data, client_configuration=configuration)

    api = client.ApiClient(configuration=configuration)

    kubeconfig_request = {
        'apiVersion': 'authentication.gardener.cloud/v1alpha1',
        'kind': 'AdminKubeconfigRequest',
        'spec': {
        'expirationSeconds': expiry_seconds
        }
    }

    response = api.call_api(resource_path=f'/apis/core.gardener.cloud/v1beta1/namespaces/{project_namespace}/shoots/{shoot_name}/adminkubeconfig',
                            method='POST',
                            body=kubeconfig_request,
                            auth_settings=['BearerToken'],
                            _preload_content=False,
                            _return_http_data_only=True,
                        )

    return base64.b64decode(json.loads(response.data)["status"]["kubeconfig"]).decode('utf-8')


class FilterModule(object):
    def filters(self):
        return {
            'shoot_admin_kubeconfig': shoot_admin_kubeconfig,
        }
```

Optionally this could also persist the kubeconfig in a file like `/tmp/kubeconfig.{{ cluster }}`.

What do you think about this?

control-plane/roles/isolated-clusters/defaults/main.yml
```diff
@@ -0,0 +1,25 @@
+---
+isolated_clusters_ntp_image_name: ghcr.io/mwennrich/chrony
+isolated_clusters_ntp_image_tag:
+isolated_clusters_ntp_namespace: ntp-service
+
+isolated_clusters_dns_image_name: coredns/coredns
+isolated_clusters_dns_image_tag:
+isolated_clusters_dns_namespace: dns-service
+
+isolated_clusters_ingress_ingress_controller_namespace: ingress
```
Gerrit91, line 10: I think to shorten names, repeating "ingress" is not really necessary, it's still clear when it's `isolated_clusters_ingress_controller_namespace`.

control-plane/roles/isolated-clusters/defaults/main.yml
```diff
@@ -0,0 +1,25 @@
+---
```
Gerrit91, line 1: You will also need to symlink the control plane defaults here as otherwise the variables from there cannot be picked up with the current repos layout.

control-plane/roles/isolated-clusters/defaults/main.yml
```diff
@@ -0,0 +1,25 @@
...
+
+isolated_clusters_dns_image_name: coredns/coredns
+isolated_clusters_dns_image_tag:
+isolated_clusters_dns_namespace: dns-service
+
+isolated_clusters_ingress_ingress_controller_namespace: ingress
+isolated_clusters_ingress_ingress_controller_chart_version:
+isolated_clusters_ingress_ingress_controller_chroot: true
+isolated_clusters_ingress_cert_manager_version:
+isolated_clusters_ingress_cert_manager_cluster_issuer: letsencrypt-gcp-dns
```
Gerrit91, line 14: Pre-suggesting the DNS provider is maybe something quite opinionated. Still not having a good idea yet but I think it would be good to give a bigger freedom to how users want to deal with where certs and ingresses come from. I will come back to this.

control-plane/roles/isolated-clusters/tasks/ingress.yaml
```diff
@@ -0,0 +1,82 @@
...
+- name: Create dns secret for gcp cluster issuer
+  k8s:
+    definition:
+      apiVersion: v1
+      kind: Secret
+      metadata:
+        name: clouddns-serviceaccount-secret
+        namespace: cert-manager
+      type: Opaque
+      data: "{{ gardener_dns_credentials }}"
```
Gerrit91, line 53: These need to become role-prefixed variables, we should not pick up any undocumented variables.

control-plane/roles/isolated-clusters/tasks/ingress.yaml
```diff
@@ -0,0 +1,82 @@
...
+      spec:
+        acme:
+          server: https://acme-v02.api.letsencrypt.org/directory
+          email: "{{ isolated_clusters_cert_manager_lets_encrypt_expiry_mail_address }}"
+          privateKeySecretRef:
+            name: letsencrypt-gcp-dns
+          solvers:
+          - dns01:
+              cloudDNS:
+                project: "{{ gcp_project }}"
```
Gerrit91, line 74: Non-role variable should not be used.

control-plane/roles/isolated-clusters/templates/dns/coredns-deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+      labels:
+        k8s-app: coredns
+    spec:
+      containers:
+      - args:
+        - -conf
+        - /etc/coredns/Corefile
+        env:
+        image: "{{ isolated_clusters_dns_image_name }}:{{ isolated_clusters_dns_image_tag }}"
+        imagePullPolicy: IfNotPresent
```
Gerrit91, line 34: There is a control plane default variable for pull policy, maybe would be nice to use it.

control-plane/roles/isolated-clusters/templates/ingress/values.yaml
```diff
@@ -0,0 +1,26 @@
+controller:
+  image:
+    chroot: {{ isolated_clusters_ingress_ingress_controller_chroot | default(false) }}
+
+  replicaCount: {{ isolated_clusters_ingress_ingress_controller_replicas | default(1) }}
+  extraArgs:
+    enable-ssl-passthrough: ""
```
Gerrit91, line 7: Is there a specific reason we need to enable SSH passthrough?

control-plane/roles/isolated-clusters/templates/registry/registry-oci-mirror.yaml
```diff
@@ -0,0 +1,167 @@
...
+  failedJobsHistoryLimit: 1
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - args:
+            - mirror
+            - --mirror-config=/config/oci-mirror.yaml
+            image: ghcr.io/metal-stack/oci-mirror:{{ isolated_clusters_registry_oci_mirror_version }}
```
Gerrit91, line 18: Factor out image name as well to allow using it from somewhere else easily.

control-plane/roles/isolated-clusters/templates/registry/registry-oci-mirror.yaml
```diff
@@ -0,0 +1,167 @@
...
+              secretName: oci-mirror-config
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: oci-mirror-config
+stringData:
+  oci-mirror.yaml: |
+      images:
+        - source: docker.io/fluent/fluentd
```
Gerrit91, line 40: I think this image list should be a default variable.

control-plane/roles/isolated-clusters/templates/registry/registry-sts.yaml
```diff
@@ -0,0 +1,54 @@
...
+  serviceName: registry
+  template:
+    metadata:
+      labels:
+        app: registry
+        name: registry
+    spec:
+      containers:
+      - name: registry
+        image: registry:{{ isolated_clusters_registry_version }}
```
Gerrit91, line 26: Factor out image name as well to allow using it from somewhere else easily.

control-plane/roles/isolated-clusters/templates/registry/registry-sts.yaml
```diff
@@ -0,0 +1,54 @@
...
+    kind: PersistentVolumeClaim
+    metadata:
+      name: storage
+    spec:
+      accessModes:
+      - ReadWriteOnce
+      resources:
+        requests:
+          storage: {{ isolated_clusters_registry_storage_size }}
+      storageClassName: partition-gold
```
Gerrit91, line 53: Should be optional and factored out as this only works for metal-stack shoots.

Gerrit91 (commented):
Thanks for this first effort to make this available for public. It's pretty tough to get this right so I hope I do not make it too difficult for you with my review comments. Please talk to me if you need help with this.

control-plane/roles/isolated-clusters/tasks/dns.yaml
```diff
@@ -0,0 +1,27 @@
...
+  k8s:
+    definition:
+      api_version: v1
+      kind: Namespace
+      metadata:
+        name: "{{ isolated_clusters_dns_namespace }}"
+        labels:
+          group: "{{ isolated_clusters_group_label }}"
+  environment:
+    K8S_AUTH_KUBECONFIG: "/tmp/kubeconfig.{{ cluster }}"
```
robertvolkmann, line 16: Please use key `kubeconfig`, because we pass the kubeconfig as a dictionary instead of a file:
```suggestion
- name: Create namespace {{ isolated_clusters_dns_namespace }}
  k8s:
    definition:
      api_version: v1
      kind: Namespace
      metadata:
        name: "{{ isolated_clusters_dns_namespace }}"
        labels:
          group: "{{ isolated_clusters_group_label }}"
    kubeconfig: "{{ kubeconfig as dictionary }}"
```

control-plane/roles/isolated-clusters/tasks/dns.yaml
robertvolkmann: Please don't persist it as a file. It is no longer necessary. You can pass the whole dictionary.

control-plane/roles/isolated-clusters/tasks/dns.yaml
```diff
@@ -0,0 +1,27 @@
...
+- name: Deploy DNS service
+  k8s:
+    definition: "{{ lookup('template', 'dns/' + item) }}"
+    namespace: "{{ isolated_clusters_dns_namespace }}"
+  loop:
+    - coredns-configmap.yaml
+    - coredns-deployment.yaml
+    - coredns-service.yaml
+  environment:
+    K8S_AUTH_KUBECONFIG: "/tmp/kubeconfig.{{ cluster }}"
```
robertvolkmann, line 27:
```suggestion
- name: Deploy DNS service
  k8s:
    kubeconfig: "/tmp/kubeconfig.{{ cluster }}"
    template:
    - path: dns/coredns-configmap.yaml
    - path: dns/coredns-deployment.yaml
    - path: dns/coredns-service.yaml

File: control-plane/roles/isolated-clusters/tasks/ingress.yaml [hunk 16]
```diff
@@ -0,0 +1,82 @@
...
+      kind: Namespace
+      metadata:
+        name: "{{ isolated_clusters_ingress_controller_namespace }}"
+    apply: yes
+  environment:
+    K8S_AUTH_KUBECONFIG: "/tmp/kubeconfig.{{ cluster }}"
+
+- name: Deploy ingress-controller
+  include_role:
+    name: ansible-common/roles/helm-chart
```
robertvolkmann, line 19: Please use community version `kubernetes.core.helm`.

control-plane/roles/isolated-clusters/templates/registry/registry-oci-mirror.yaml
```diff
@@ -0,0 +1,167 @@
...
+          - name: oci-mirror-config
+            secret:
+              defaultMode: 420
+              items:
+              - key: oci-mirror.yaml
+                path: oci-mirror.yaml
+              secretName: oci-mirror-config
+---
+apiVersion: v1
+kind: Secret
```
robertvolkmann, line 34: Should be a ConfigMap because it doesn't contain any secrets.

control-plane/roles/isolated-clusters/templates/registry/registry-ingress.yaml
```diff
@@ -0,0 +1,33 @@
...
+  ingressClassName: nginx
+  rules:
+  - host: "{{ isolated_clusters_registry_ingress_fqdn }}"
+    http:
+      paths:
+      - backend:
+          service:
+            name: registry
+            port:
+              number: 5000
```
robertvolkmann, line 27:
```suggestion
              name: registry
```

control-plane/roles/isolated-clusters/templates/registry/registry-service.yaml
```diff
@@ -0,0 +1,13 @@
...
+  labels:
+    app: registry
+  name: registry
+spec:
+  selector:
+    app: registry
+  ports:
+  - name: registry
+    port: 5000
+    targetPort: registry
```
robertvolkmann, line 13:
```suggestion
    targetPort: http
```

control-plane/roles/isolated-clusters/templates/registry/registry-sts.yaml
```diff
@@ -0,0 +1,54 @@
...
+        image: registry:{{ isolated_clusters_registry_version }}
+        imagePullPolicy: IfNotPresent
+        env:
+        - name: REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY
+          value: "/var/lib/registry"
+        - name: REGISTRY_STORAGE_DELETE_ENABLED
+          value: "true"
+        ports:
+        - containerPort: 5000
+          name: registry
```
robertvolkmann, line 35:
```suggestion
          name: http
```

control-plane/roles/isolated-clusters/templates/registry/registry-sts.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: registry
```
robertvolkmann, line 4: Why do we need a ServiceAccount?

control-plane/roles/isolated-clusters/templates/ntp/chrony-deployment.yaml
```diff
@@ -0,0 +1,94 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: chrony
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    k8s-app: chrony
```
robertvolkmann, line 10: Please use the recommended labels, see https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels
```suggestion
    app.kubernetes.io/name: chrony
```

control-plane/roles/isolated-clusters/tasks/ingress.yaml
```diff
@@ -0,0 +1,82 @@
...
+    helm_repo: https://kubernetes.github.io/ingress-nginx
+    helm_chart_version: "{{ isolated_clusters_ingress_controller_chart_version }}"
+    helm_chart: ingress-nginx
+    helm_release_name: ingress-nginx
+    helm_target_namespace: "{{ isolated_clusters_ingress_controller_namespace }}"
+    helm_value_file_template: ingress/values.yaml
+
+- name: Deploy Cert Manager for Registry
+  include_role:
+    name: cert-manager
```
robertvolkmann, line 33: This role doesn't exist in metal-roles.

robertvolkmann:
Why not creating a Helm Chart and only consume it here?

control-plane/roles/isolated-clusters/defaults/main.yml
vknabel: Drop cert manager (part of actual deployment of user)

control-plane/roles/isolated-clusters/templates/ingress/values.yaml
vknabel: Can probably be deleted

control-plane/roles/isolated-clusters/templates/registry/registry-sts.yaml
vknabel: Introduce: `isolated_clusters_registry_storage_class_name`

control-plane/roles/isolated-clusters/templates/registry/registry-sts.yaml
vknabel: Not needed :+1:

control-plane/roles/isolated-clusters/templates/registry/registry-ingress.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  annotations:
+    {% for key, value in isolated_clusters_registry_ingress_annotations.items() %}
+        {{ key }}: "{{ value }}"
+    {% endfor %}
```
robertvolkmann, line 7:
```suggestion
      {{ isolated_clusters_registry_ingress_annotations | indent(width=4, first=false) }}
```

control-plane/roles/isolated-clusters/tasks/dns.yaml
Gerrit91: Revert beause:

```
  msg: 'Failed to create object: The `items` field in the body must be populated when calling methods on a ResourceList'
``` 

Maybe need to look at this one more time.

Gerrit91 (approved):

PR #274, opened 2024-03-19
Support new GPU constraint type which does not include min and max.

majst01:


Gerrit91:
Not required anymore because there are `min` and `max` for GPU constraints as well.

PR #278, opened 2024-04-23
L2 leaves

mreiger:
Add variables for creating a leaf pair that has a Layer-2 VLAN and connects machines through MCLAG

### DIFF
--- a/partition/roles/sonic/templates/frr.conf.j2
+++ b/partition/roles/sonic/templates/frr.conf.j2
@@ -18,14 +18,6 @@ vrf {{ i.vrf }}
 exit-vrf
 {% endif %}
 {% endfor %}
-{% if sonic_vrfs is defined and sonic_vrfs|length > 0 %}
-{% for vrf in sonic_vrfs %}
-!
-vrf Vrf{{ vrf.id }}
- vni {{ vrf.vni }}
-exit-vrf
-{% endfor %}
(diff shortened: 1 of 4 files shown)


partition/roles/sonic/templates/frr.conf.j2
```diff
@@ -18,6 +18,14 @@ vrf {{ i.vrf }}
...
 {% endif %}
 {% endfor %}
+{% if sonic_vrfs is defined and sonic_vrfs|length > 0 %}
+{% for vrf in sonic_vrfs %}
+!
+vrf Vrf{{ vrf.id }}
+ vni {{ vrf.vni }}
+exit-vrf
+{% endfor %}
+{% endif %}
```
robertvolkmann, line 28: What is the difference to `sonic_interconnects`?

robertvolkmann (changes requested):
The tests still fails.

partition/roles/sonic/templates/frr.conf.j2
mreiger: This was left over from an earlier configuration experiment. I have deleted it for now; if something like this is necessary later on it can be recreated easily enough.

robertvolkmann (approved):

Gerrit91 (approved):

PR #281, opened 2024-04-29
Move `leaf` and `router` role from metal-roles into mini-lab.

Gerrit91:
References https://github.com/metal-stack/mini-lab/issues/155.

```ACTIONS_REQUIRED
The `leaf` and `router` role were removed from the partition deployment roles because we think only the mini-lab used them. Please make sure not to rely on these roles anymore in your automations.
```


robertvolkmann (changes requested):
There are still references in partition/Readme.md

robertvolkmann (approved):

PR #284, opened 2024-06-06
Provide default for sonic portchannels.

Gerrit91:


partition/roles/sonic/defaults/main.yaml
```diff
@@ -11,9 +11,11 @@ sonic_ports: []
 sonic_ports_dict: {}
 sonic_ports_default_fec: none
 
+sonic_portchannels: []
```
robertvolkmann, line 14: [Task](https://github.com/metal-stack/metal-roles/pull/278/files#diff-1338257ac65a640ff4601ce33a2d2211bfb7fb8585f87766d5e3176f3f39c585) has the wrong condition and should instead use:
`sonic_portchannels is defined and sonic_portchannels|length > 0 `

Gerrit91: The other tasks similar to this also just use the same condition. It should give same results when having this variable set as the default.

mreiger (approved):

PR #286, opened 2024-06-21
Allow configuring members of the gardener soil project

iljarotar:


control-plane/roles/gardener/tasks/main.yaml
```diff
@@ -123,7 +123,7 @@
   k8s:
     definition: "{{ lookup('template', 'gardener-soil-project.yaml.j2') }}"
     kubeconfig: "{{ gardener_kube_apiserver_kubeconfig_path }}"
-  when: not lookup('k8s', kubeconfig=gardener_kube_apiserver_kubeconfig_path, api_version='core.gardener.cloud/v1beta1', kind='Project', resource_name=gardener_soil_name)
```
robertvolkmann, line 153: @Gerrit91 Do you remember why this was necessary?

Gerrit91: Not sure. Maybe the `metadata.namespace` field was an issue, which is a bit special for this resource. If it works with apply without creating multiple project resources I think we can make this change.

Gerrit91: The project is not a namespaced resource so when the name is given it will not create more projects.

Gerrit91 (approved):

Issue #285, opened 2024-06-20
Docker Logs might consume all disk space on Switches


vknabel:
After some time running the `grafana/promtail` container logs consume all free space on the disk ultimately resulting in failing switch syncs.
It would make sense to limit the amount of logs to prevent this.

majst01:
Can you identify what exactly must be done to restrict the log volume ?

vknabel:
Not sure how, but we need to set the options in https://docs.docker.com/config/containers/logging/json-file/

majst01:
> Not sure how, but we need to set the options in https://docs.docker.com/config/containers/logging/json-file/

This applies to all other containers on the switch then, maybe there is some SONiC documentation on this topic

majst01:
Be aware that this affects all switches and management servers with the consequence that all daemons which are started with docker will be restarted

vknabel:
With the solution all daemons started by the `systemd-docker-service` role will be restarted. If understand it correctly this does not apply sonic's daemons only 'ours'.

PR #290, opened 2024-07-12
fix(sonic): docker log size

vknabel:
for #285 (cc @iljarotar )


partition/roles/sonic/tasks/main.yaml
```diff
@@ -42,6 +42,11 @@
     src: resolv.conf.j2
     dest: /etc/resolv.conf
 
+- name: render docker daemon.json
```
majst01, line 45: But then a handler must be triggered to restart the dockerd

vknabel: I am unsure if the changes in [the latest commit](https://github.com/metal-stack/metal-roles/pull/290/commits/4d10c3e055e811f2bf23c6a4c3c68a0c2b4dd6f6) are correct. Can you have another look?

robertvolkmann (changes requested):
Please don't change the `daemon.json`. Every container originated from SONiC already sets:
```
"LogConfig": {
  "Type": "json-file",
  "Config": {
    "max-file": "5",
    "max-size": "2M"
  }
}
```
Please use the same settings for our containers and everything is fine, e.g.
```
--log-opt max-size=2M --log-opt max-file=5
```
The mgmt servers also miss any limits.

robertvolkmann:
Could you also remove this nasty default for metal-core?
https://github.com/metal-stack/metal-roles/blob/master/partition/roles/metal-core/defaults/main/main.yaml#L2

Gerrit91:
Suggestion from @robertvolkmann was also to extend the [systemd-docker-service](https://github.com/metal-stack/ansible-common/tree/master/roles/systemd-docker-service) role to include two new template variables for configuring `--log-opt max-size` and `--log-opt max-file` along with the existing `--log-driver` flag (maybe even set reasonable default in case log driver is json-file in order to prevent disks running full on the mgmt-servers for example).

The advantage would also be that we do not need to restart the Dockerd during the rollout.

What do you think?

vknabel:
Applied your suggestions in https://github.com/metal-stack/ansible-common/pull/30
If `systemd_docker_log_driver` is not set, `json-file` will be the default. And in this case we limit the log files and size. So nothing to do here anymore.

vknabel:
not required anymore

PR #292, opened 2024-07-24
Support mgmt-servers that do not use systemd-resolved

iljarotar:


partition/roles/dhcp/templates/dhcpd.conf.j2
```diff
@@ -4,6 +4,8 @@ authoritative;
 default-lease-time {{ dhcp_default_lease_time }};
 max-lease-time {{ dhcp_max_lease_time }};
 
+option domain-name-servers {{ mgmt_server_nameservers |join(',') }};
```
Gerrit91, line 7: Please make this optional to maintain backwards compatibility and use variables documented with the role name (`dhcp_`) and document the usage in the role's README.

Gerrit91 (approved):

PR #300, opened 2024-07-31
Allow different ssh key file names than the default id_rsa

iljarotar:
Also added new docker installation to this PR so all of the changes can be tested with an image override. Will be split afterwards.

### DIFF
--- a/partition/roles/mgmt-server/tasks/main.yaml
+++ b/partition/roles/mgmt-server/tasks/main.yaml
@@ -39,11 +39,26 @@
     name:
     - net-tools
     - ipmitool
-    - docker.io
     - iptables-persistent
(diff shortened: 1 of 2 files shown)


partition/roles/mgmt-server/tasks/main.yaml
```diff
@@ -39,11 +39,26 @@
     name:
     - net-tools
     - ipmitool
-    - docker.io
```
Gerrit91, line 42: Unrelated changes, cannot approve. :sweat_smile:

iljarotar: removed

Gerrit91 (approved):

PR #301, opened 2024-08-01
Install docker from official repository

iljarotar:


partition/roles/mgmt-server/tasks/main.yaml
```diff
@@ -39,11 +39,26 @@
     name:
     - net-tools
     - ipmitool
-    - docker.io
     - iptables-persistent
     update_cache : yes
     force_apt_get: yes
 
+- name: install docker
```
Gerrit91, line 46: Unfortunately this is breaking for existing management server deployments.

One idea would be to figure out if docker is already installed an then skip this step.

Gerrit91 (approved):

PR #309, opened 2024-08-14
Sag gratuitous arp override

mreiger:
Set the option to accept gratuitous arp on Vlans with SAG enabled as recommended by Edgecore; see https://support.edge-core.com/hc/en-us/articles/6840823990169--Enterprise-SONiC-SAG-static-anycast-gateway

Only implemented for VLANs in a VRF since this is the current use case.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 .shoot*
 __pycache__
 .Trash*
+*~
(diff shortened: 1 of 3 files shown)


mreiger:
Tested in integration environment

vknabel (approved):

majst01 (approved):

PR #315, opened 2024-09-10
Allow to configure email and server of cert management issuer

simcod:
## Required Actions

```ACTIONS_REQUIRED
Operators need to define the new variable `gardener_cert_management_issuer_email` before deploying Gardener.
```

### DIFF
--- a/control-plane/roles/gardener/tasks/extensions.yaml
+++ b/control-plane/roles/gardener/tasks/extensions.yaml
@@ -69,9 +69,7 @@
   loop:
     - controller-deployment.yaml
     - controller-registration.yaml
-  when:
-    - gardener_extension_networking_cilium_enabled
-    - gardener_cert_management_issuer_email is defined
+  when: gardener_extension_networking_cilium_enabled
 
 - name: "Register controller: shoot-cert-service"
   k8s:
--- a/control-plane/roles/gardener/tasks/extensions.yaml
+++ b/control-plane/roles/gardener/tasks/extensions.yaml
@@ -69,7 +69,9 @@
   loop:
     - controller-deployment.yaml
     - controller-registration.yaml
-  when: gardener_extension_networking_cilium_enabled
+  when:
+    - gardener_extension_networking_cilium_enabled
+    - gardener_cert_management_issuer_email is defined
 
 - name: "Register controller: shoot-cert-service"
   k8s:
--- a/control-plane/roles/gardener/tasks/main.yaml
+++ b/control-plane/roles/gardener/tasks/main.yaml
@@ -55,6 +55,7 @@
       - gardener_dns_provider is not none
       - gardener_cloud_profile_metal_api_url is not none
       - gardener_cloud_profile_metal_api_hmac is not none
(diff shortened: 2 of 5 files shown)


control-plane/roles/gardener/templates/shoot-cert-service/controller-deployment.yaml
```diff
@@ -15,7 +15,7 @@ providerConfig:
       defaultIssuer:
         restricted: true # restrict default issuer to any sub-domain of shoot.spec.dns.domain
         acme:
-          email: cert-expiry@metal-pod.io
-          server: https://acme-v02.api.letsencrypt.org/directory
+          email: "{{ gardener_cert_management_issuer_email }}"
+          server: "{{ gardener_cert_management_issuer_server }}"
```
robertvolkmann, line 19: Missing documentation in README.md

robertvolkmann (changes requested):

control-plane/roles/gardener/templates/shoot-cert-service/controller-deployment.yaml
Gerrit91, line 19: Can we default this to value to `https://acme-v02.api.letsencrypt.org/directory` to reduce required actions for operators?

Gerrit91: Also looks like deployment would fail without having these variables defined. So we need to make the email required and should assert that it is defined.

control-plane/roles/gardener/tasks/extensions.yaml
```diff
@@ -69,7 +69,9 @@
   loop:
     - controller-deployment.yaml
     - controller-registration.yaml
-  when: gardener_extension_networking_cilium_enabled
+  when:
+    - gardener_extension_networking_cilium_enabled
+    - gardener_cert_management_issuer_email is defined
```
Gerrit91, line 74: I think this is the wrong place? I also meant adding an assert, which is at the beginning of tasks of this role in the `main.yaml`. The task is called `Check mandatory variables for this role are set`.

simcod: Now it should be in the right place, I think.

control-plane/roles/gardener/tasks/main.yaml
```diff
@@ -55,6 +55,7 @@
       - gardener_dns_provider is not none
       - gardener_cloud_profile_metal_api_url is not none
       - gardener_cloud_profile_metal_api_hmac is not none
+      - gardener_cert_management_issuer_email is not none
```
Gerrit91, line 58: Yes, place is now correct, however when checking `is not none` you should add `gardener_cert_management_issuer_email:` to the defaults. Otherwise Ansible will fail with "variable not defined" but not with the assertion.

It would also be possible to check for `is defined`. But I think first variant is little better.

Gerrit91 (approved):

robertvolkmann (approved):

PR #316, opened 2024-09-10
Make basic_auth for prometheus and promtail optional

simcod:


Gerrit91 (approved):
As there's no documentation at all for this role, we can simply merge it. Would be nice to have some docs though for the future.

partition/roles/monitoring/prometheus/templates/prometheus.yaml.j2
```diff
@@ -34,9 +34,11 @@ alerting:
 {% if prometheus_remote_write_url %}
 remote_write:
 - url: {{ prometheus_remote_write_url }}
+  {% if prometheus_remote_write_basic_auth_username and prometheus_remote_write_basic_auth_password %}
```
robertvolkmann, line 37:
```suggestion
  {% if prometheus_remote_write_basic_auth_username is defined and prometheus_remote_write_basic_auth_password is defined %}
```

partition/roles/promtail/templates/promtail.yaml.j2
```diff
@@ -8,9 +8,11 @@ positions:
 clients:
   - url: {{ promtail_loki_push_endpoint }}
     timeout: 60s
+    {% if promtail_loki_basic_auth_username and promtail_loki_basic_auth_password %}
```
robertvolkmann, line 11:
```suggestion
    {% if promtail_loki_basic_auth_username is defined and promtail_loki_basic_auth_password is defined %}
```

robertvolkmann (changes requested):

PR #318, opened 2024-09-10
Allow to provide the imageVectorOverwrite for the cilium extension

simcod:
Alternative suggestions for the variable name are welcome.

## Breaking Change

```BREAKING_CHANGE
The static Cilium version image vector overwrite was removed. The extension now gets deployed by the image vector shipped with the networking extension.
```

### DIFF
--- a/control-plane/roles/gardener/defaults/main/extensions.yaml
+++ b/control-plane/roles/gardener/defaults/main/extensions.yaml
@@ -1,6 +1,7 @@
 ---
 gardener_extension_networking_calico_enabled: true
 gardener_extension_networking_cilium_enabled: true
+gardener_extension_networking_cilium_vector_overwrite_image_tag: v1.12.1
 gardener_extension_os_metal_enabled: true
 gardener_extension_provider_gcp_enabled: true
 gardener_extension_provider_metal_enabled: true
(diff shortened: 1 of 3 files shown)


control-plane/roles/gardener/defaults/main/extensions.yaml
```diff
@@ -1,6 +1,7 @@
 ---
 gardener_extension_networking_calico_enabled: true
 gardener_extension_networking_cilium_enabled: true
+gardener_extension_networking_cilium_vector_overwrite_image_tag: v1.12.1
```
Gerrit91, line 4: To reduce this maintenance overhead I propose adding the variable `gardener_extension_networking_cilium_image_vector_overwrite`. By default this is empty and no overwrite is being made. This way the extension just deploys the image vector bundled together with the extension. Operators can provide an overwrite completely on their own.

We will declare this a a breaking change.

control-plane/roles/gardener/templates/networking-cilium/controller-deployment.yaml
```diff
@@ -11,21 +11,10 @@ providerConfig:
...
-          tag: v1.12.1
-        - name: cilium-operator
-          sourceRepository: github.com/cilium/cilium
-          repository: quay.io/cilium/operator
-          tag: v1.12.1
-        - name: hubble-relay
-          sourceRepository: github.com/cilium/hubble-ui
-          repository: quay.io/cilium/hubble-relay
-          tag: v1.12.1
+{% for imageOverwrite in gardener_extension_networking_cilium_image_vector_overwrite %}
```
Gerrit91, line 17: Use `to_nice_yaml` filter to produce valid stuff here.

Gerrit91 (approved):

PR #320, opened 2024-09-23
use default speed, mtu and fec for unconfigured ports as well

iljarotar:


partition/roles/sonic/templates/metal.yaml.j2
```diff
@@ -114,7 +114,9 @@ PORT:
     mtu: "{{ port.mtu|default(sonic_ports_default_mtu) }}"
     fec: "{{ port.fec|default(sonic_ports_default_fec)|string|lower }}"
     {% else %}
-    speed: "{{ running_cfg.speed }}"
+    speed: "{{ sonic_ports_default_speed|default(running_cfg.speed) }}"
+    mtu: "{{ sonic_ports_default_mtu|default(running_cfg.mtu) }}"
```
robertvolkmann, line 118: metal-core is also setting MTU. How do you want to resolve this conflict?

robertvolkmann (changes requested):

iljarotar:
As pointed out, metal-core already sets mtu. So no need for this PR.

PR #323, opened 2024-09-26
Remove sonic ports defaults

iljarotar:
## Additional Description

Currently, if the `sonic_ports` variable is defined the sonic role will require default values for fec, mtu and speed to be set. A more reasonable behavior, in my opinion, is to leave all settings of the running configuration in place, if no value was provided.

## Breaking Change

```BREAKING_CHANGE
`sonic_ports_default_mtu`, `sonic_ports_default_fec` and `sonic_ports_default_speed` are not used anymore. Each port requires and individual configuration.
```

### DIFF
--- a/partition/roles/sonic/templates/metal.yaml.j2
+++ b/partition/roles/sonic/templates/metal.yaml.j2
@@ -108,15 +108,15 @@ PORT:
     parent_port: {{ running_cfg.parent_port }}
     {% endif %}
     admin_status: up
-    {% set port = sonic_ports_dict[name] %}
-    {% if (port is defined and port.speed is defined) or running_cfg.speed is defined %}
-    speed: "{{ (port | default({})).speed | default(running_cfg.speed) }}"
+    {% set port = sonic_ports_dict[name] | default({}) %}
+    {% if port.speed is defined or running_cfg.speed is defined %}
+    speed: "{{ port.speed | default(running_cfg.speed) }}"
     {% endif %}
-    {% if (port is defined and port.mtu is defined) or running_cfg.mtu is defined %}
-    mtu: "{{ (port | default({})).mtu | default(running_cfg.mtu) }}"
+    {% if port.mtu is defined or running_cfg.mtu is defined %}
+    mtu: "{{ port.mtu | default(running_cfg.mtu) }}"
     {% endif %}
-    {% if (port is defined and port.fec is defined) or running_cfg.fec is defined %}
-    fec: "{{ (port | default({})).fec | default(running_cfg.fec) }}"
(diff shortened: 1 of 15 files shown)


partition/roles/sonic/templates/metal.yaml.j2
```diff
@@ -107,14 +107,16 @@ PORT:
...
-    speed: "{{ running_cfg.speed }}"
+    {% set port = sonic_ports_dict[name] %}
+    {% if (port is defined and port.speed is defined) or running_cfg.speed is defined %}
+    speed: "{{ (port | default({})).speed | default(running_cfg.speed) }}"
+    {% endif %}
+    {% if (port is defined and port.mtu is defined) or running_cfg.mtu is defined %}
+    mtu: "{{ (port | default({})).mtu | default(running_cfg.mtu) }}"
+    {% endif %}
+    {% if (port is defined and port.fec is defined) or running_cfg.fec is defined %}
+    fec: "{{ (port | default({})).fec | default(running_cfg.fec) }}"
```
robertvolkmann, line 119: Could you try?
```suggestion
    {% set port = sonic_ports_dict.get(name, {}) %}
    {% if port.speed is defined or running_cfg.speed is defined %}
    speed: "{{ (port.speed | default(running_cfg.speed) }}"
    {% endif %}
    {% if port.mtu is defined or running_cfg.mtu is defined %}
    mtu: "{{ (port.mtu | default(running_cfg.mtu) }}"
    {% endif %}
    {% if port.fec is defined or running_cfg.fec is defined %}
    fec: "{{ (port.fec | default(running_cfg.fec) }}"
```

partition/roles/sonic/templates/metal.yaml.j2
```diff
@@ -107,14 +107,16 @@ PORT:
     {% if 'parent_port' in running_cfg %}
     parent_port: {{ running_cfg.parent_port }}
     {% endif %}
-    {% if sonic_ports_dict[name] is defined %}
-    {% set port = sonic_ports_dict[name] %}
     admin_status: up
```
robertvolkmann, line 110: Why always setting admin_status to up? Why not similar to fec, mtu, and speed?

robertvolkmann (changes requested):
Is setting fec, mtu, and speed still tested?

partition/roles/sonic/templates/metal.yaml.j2
iljarotar: We could introduce another variable for the admin_status, but that would possibly involve a lot of changes to the existing deployments, right?

Gerrit91:
Closing in favor of #324.

PR #324, opened 2024-09-27
Use running configuration for sonic ports

iljarotar:
## References

Replaces https://github.com/metal-stack/metal-roles/pull/323

## Additional Description

This is a compromise that doesn't introduce any breaking changes. Default mtu, speed and fec for SONiC ports can still be used. Additionally, if the running configuration holds values for mtu, speed and fec these values will be considered for the new config. The reasoning behind this PR is that all values that are not explicitly defined should remain as they are, while the current template will remove mtu and fec from ports, that are not part of the `sonic_ports` dictionary.

### DIFF
--- a/partition/roles/sonic/templates/metal.yaml.j2
+++ b/partition/roles/sonic/templates/metal.yaml.j2
@@ -107,13 +107,14 @@ PORT:
     {% if 'parent_port' in running_cfg %}
     parent_port: {{ running_cfg.parent_port }}
     {% endif %}
-    admin_status: up
     {% if sonic_ports_dict[name] is defined %}
     {% set port = sonic_ports_dict[name] %}
+    admin_status: {{ port.admin_status|default('up') }}
     speed: "{{ port.speed|default(sonic_ports_default_speed) }}"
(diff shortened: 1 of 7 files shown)


partition/roles/sonic/templates/metal.yaml.j2
```diff
@@ -107,15 +107,23 @@ PORT:
     {% if 'parent_port' in running_cfg %}
     parent_port: {{ running_cfg.parent_port }}
     {% endif %}
+    admin_status: up
     {% if sonic_ports_dict[name] is defined %}
     {% set port = sonic_ports_dict[name] %}
-    admin_status: up
```
robertvolkmann, line 112: Why was this changed?

iljarotar: To make deployments less verbose. Otherwise you have to explicitly add every port to the `sonic_ports_dict`, where you want to have `admin_status: up`.

robertvolkmann: Could you allow to deactivate a port?

robertvolkmann (approved):

PR #325, opened 2024-09-27
Restart bgp after initial metal-core deployment

iljarotar:
If metal-core is deployed for the first time on a switch, the bgp session does not come up before bgp is restarted. Currently this is done manually. This PR is an attempt to automate this process. It needs some more testing to see if it really solves the problem.


mreiger (approved):

iljarotar:
Need to fix the underlying issue instead of using this workaround.

PR #327, opened 2024-10-04
Added a role to boot a SONiC switch to onie

mreiger:
## Release Notes

A new role was added to boot a SONiC switch into ONIE. It is useful for reinstalling a switch from scratch without upgrading to a different SONiC version.


partition/roles/sonic-boot-to-onie/tasks/main.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - metal_stack_switch_os_is_sonic
+
+- name: Set next boot entry to ONIE
+  command: efibootmgr -n {{ sonic_boot_to_onie_bootentry }}
+
+- name: Boot into ONIE
+  reboot:
+
+- name: Wait until ZTP has completed
+  shell: "show ztp status"
```
Gerrit91, line 20: Is there a specific reason to use `shell` instead of  `command`? In general, `command` should be preferred.

mreiger: Fixed.

mreiger:
Should we add the option to skip waiting for ZTP, in case someone does not use it, or is it safe to assume anyone using metal-roles is doing ZTP?

Gerrit91:
I think operators can decide not to add this role to their playbooks if the do not want to use it?

mreiger:
Naturally. :D
I was wondering if I should add a switch, in case you do want to reinstall your switch from a playbook but don't use ZTP.
Anyway; can be added later if someone actually wants this.

mreiger:
Note to self: The boot entry is not always the same even on the same switch model; need to parse the output of `efibootmgr` to make this role stable.

iljarotar:
Closing this as stale. If still needed please re-open.

PR #331, opened 2024-10-11
feat(monitoring): allow disabling prometheus operator

vknabel:
## Additional Description

In some cases CRDs and the prometheus operator will already be installed by Gardener itself leading to two conflicting instances of the operator. 
These changes allow disabling deploying the prometheus operator to solve this conflict. We also rely on the helm-chart to deploy the CRDs from now on, which is enabled by default and handles existing CRDs better.

## Release Notes

- introduces `monitoring_prometheus_operator_enabled` to be able to prevent conflicts with the Gardener Prometheus Operator.  Defaults to true to retain the old behavior.


vknabel:
**Info:** currently untested

control-plane/roles/monitoring/tasks/main.yaml
```diff
@@ -27,6 +27,7 @@
 - name: Determine app version for prometheus operator CRDs
   command: helm show chart --repo {{ prometheus_stack_repo }} --version {{ prometheus_chart_version }} kube-prometheus-stack
   register: helm_output
+  when: monitoring_crds_enabled
```
simcod, line 30: Should this not rather be: `monitoring_prometheus_operator_crds_enabled`

control-plane/roles/monitoring/tasks/main.yaml
```diff
@@ -40,6 +41,7 @@
     - monitoring.coreos.com_prometheusrules.yaml
     - monitoring.coreos.com_servicemonitors.yaml
     - monitoring.coreos.com_thanosrulers.yaml
+  when: monitoring_crds_enabled
```
simcod, line 44: Same here. Should this not rather be: `monitoring_prometheus_operator_crds_enabled`

robertvolkmann:
@vknabel, are you planning to upgrade the kube-prometheus helm chart to at least version 59.1.0 to ensure compatibility with prometheus-operator v0.74.0, which is deployed by Gardener 1.99?

vknabel:
A dependency upgrade of the kube-prometheus helm chart should be done in a different PR

ma-hartma:
@simcod and me set up the monitoring on the demo-rack, running into the same issue.
Not all environments will run into this, but I don't see why we shouldn't add these flags if they work. 
Is it fine if I rebase the branch and test it @vknabel? :)

vknabel:
sure, please go ahead :)

control-plane/roles/monitoring/templates/prometheus-stack-values.yaml
```diff
@@ -1,4 +1,7 @@
 ---
+crds:
+  enabled: {{ monitoring_prometheus_operator_crds_enabled }}
```
Gerrit91: Probably we can even disable this because they are historically deployed in a task prior to deploying the helm-chart? On the other hand we can also switch to this mechanism if this works now (have no experience with it).

ma-hartma:
This was tested on the demo-rack and works fine. Ready to go imo.

ma-hartma (approved):

simcod (approved):

Gerrit91:
@ma-hartma But what about my comment? Aren't the CRDs now deployed twice? Shouldn't we just deploy them once?

ma-hartma:
@Gerrit91 it depends on the setup. When using the same cluster for hosting gardener and the control-plane, the operator is already deployed by the gardener setup. This flags now allow to NOT re-deploy it. An example is the demo-rack, where an operator already exists, leading to errors in the cluster on every monitoring re-deploy. 
Of course we could disable the operator CRDs per default, but that may have consequences for existing users.

As the recommendation is to use different clusters for control-plane and gardener - I think that keeping them enabled per default does make sense.

Gerrit91:
That's not directly what I mean. We now have:

```yaml
---
crds:
  enabled: true
```

In the changes. But before applying the chart we do:

```yaml
- name: Apply prometheus operator CRDs
  kubernetes.core.k8s:
    definition: "{{ lookup('url', 'https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/'+ (helm_output.stdout | from_yaml).get('appVersion')
      + '/example/prometheus-operator-crd/' + item , split_lines=False) | regex_replace('- =\n', '- \"=\"\n') }}"
  loop:
    - monitoring.coreos.com_alertmanagerconfigs.yaml
    - monitoring.coreos.com_alertmanagers.yaml
    - monitoring.coreos.com_podmonitors.yaml
    - monitoring.coreos.com_probes.yaml
    - monitoring.coreos.com_prometheusagents.yaml
    - monitoring.coreos.com_prometheuses.yaml
    - monitoring.coreos.com_prometheusrules.yaml
    - monitoring.coreos.com_scrapeconfigs.yaml
    - monitoring.coreos.com_servicemonitors.yaml
    - monitoring.coreos.com_thanosrulers.yaml
  when: monitoring_prometheus_operator_crds_enabled
```

Why is the task still necessary if the helm chart applies the CRDs now?

ma-hartma:
Now I got you & agree! ;) Thought there was a reason for it to exist, but it is unnecessary if the helm chart manages this now.

ma-hartma:
So for our use-case it would even be enough, to leave the `crds: enabled: true` (default) and just disable the operator with `monitoring_prometheus_operator_enabled: false` - I just tested it. 
Should we even remove the `monitoring_prometheus_operator_crds_enabled` option or should we leave it as is? 
We also do not have a use-case for `monitoring_prometheus_operator_instance_namespaces`, so I removed it as well.

Gerrit91:
Just decide freely with what you want to work with, you're the maintainer. :)

PR #336, opened 2024-10-15
Add role for partition management firewall setup

Honigeintopf:
## Included

This role setups a teltonica router, automatically with the values given.
An example.yaml is given, this needs to be updated with the values that go into the firewall.

With this role it's possible to quickly setup a teltonica router, and also save the config, being able to see afterwards how a router is configured without manually ssh to it for example.

Some values are hardcoded, like turning off wireless, sms utils.


Honigeintopf:
Don't merge this yet, I will add comments and update the Readme.md. There was no don't merge label.

partition/roles/mgmt-firewall/README.md
```diff
@@ -0,0 +1,120 @@
...
+| Field          | Description    |
+| -------------- | -------------- |
+| `name`         | Interface name |
+| `ipaddr`       | IP address     |
+| `netmask`      | Subnet mask    |
+| `device`       | Router port    |
+| `dhcp_options` | (LAN Only)     |
+| `metric`       | (WAN Only)     |
+| `gateway`      | (WAN Only)     |
+| `dns` (List)   | (WAN Only)     |
```
Gerrit91, line 77: We have a dotted-syntax for dictionary fields. Maybe you can take a look somwhere here?

https://github.com/metal-stack/metal-roles/blob/master/partition/roles/dhcp/README.md?plain=1#L12-L16

Honigeintopf: These are dic fields under mgmt_firewall_lan.
What exactly do you mean?

```yaml
mgmt_firewall_lan:
    - name: ''
      ipaddr: ''
      netmask: ''
      device: ''
      dhcp_options:
        - { option: '3', value: '' }
        - { option: '6', value: '' }
        - { option: '12', value: '' }
 ```

Gerrit91: Yes, can you put this into the variables table with `mgmt_firewall_lan.name`, `mgmt_firewall_lan.ipaddr`? Then we have it similar across our roles.

PR #337, opened 2024-10-16
Remove dns-controller-manager subcomponent from release vector

Gerrit91:
## Description

The shoot-dns extension ships with the proper version of this component. In order not to have to maintain this dependency anymore, we remove the override from our release vector. Instead an operator can now manually provide an overwrite.

### Required Actions

```ACTIONS_REQUIRED
The deployment for the shoot-dns-service extension does not use the overwrite from our release vector for the dns-controller-manager subcomponent anymore. Operators who use a custom overwrite for this component now need to provide it through the new variables `gardener_shoot_dns_service_dns_controller_manager_image_name` and `gardener_shoot_dns_service_dns_controller_manager_image_tag`.
```

### DIFF
--- a/control-plane/roles/gardener/templates/shoot-dns-service/controller-deployment.yaml
+++ b/control-plane/roles/gardener/templates/shoot-dns-service/controller-deployment.yaml
@@ -11,8 +11,8 @@ providerConfig:
       repository: "{{ gardener_shoot_dns_service_image_name }}"
       tag: "{{ gardener_shoot_dns_service_image_tag }}"
 {% if gardener_shoot_dns_service_image_vector_overwrite %}
-      imageVectorOverwrite:
-        images: "{{ gardener_shoot_dns_service_image_vector_overwrite | to_json }}"
+    imageVectorOverwrite:
+      images: "{{ gardener_shoot_dns_service_image_vector_overwrite | to_json }}"
 {% endif %}
     dnsProviderManagement:
       enabled: true
--- a/control-plane/roles/gardener/templates/shoot-dns-service/controller-deployment.yaml
+++ b/control-plane/roles/gardener/templates/shoot-dns-service/controller-deployment.yaml
@@ -11,7 +11,7 @@ providerConfig:
       repository: "{{ gardener_shoot_dns_service_image_name }}"
       tag: "{{ gardener_shoot_dns_service_image_tag }}"
 {% if gardener_shoot_dns_service_image_vector_overwrite %}
-    imageVectorOverwrite:
+    imageVectorOverwrite: |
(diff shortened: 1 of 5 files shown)


control-plane/roles/gardener/templates/shoot-dns-service/controller-deployment.yaml
```diff
@@ -10,10 +10,11 @@ providerConfig:
     image:
       repository: "{{ gardener_shoot_dns_service_image_name }}"
       tag: "{{ gardener_shoot_dns_service_image_tag }}"
+{% if gardener_shoot_dns_service_image_vector_overwrite %}
+    imageVectorOverwrite:
+      images: "{{ gardener_shoot_dns_service_image_vector_overwrite | to_json }}"
```
robertvolkmann, line 15: Why not like `network-cilium`?
```suggestion
      images:
        "{{ gardener_shoot_dns_service_image_vector_overwrite | to_nice_yaml(indent=2) | indent(width=8, first=false }}"
```

control-plane/roles/gardener/templates/shoot-dns-service/controller-deployment.yaml
```diff
@@ -10,10 +10,11 @@ providerConfig:
     image:
       repository: "{{ gardener_shoot_dns_service_image_name }}"
       tag: "{{ gardener_shoot_dns_service_image_tag }}"
+{% if gardener_shoot_dns_service_image_vector_overwrite %}
+    imageVectorOverwrite:
```
robertvolkmann, line 14: According https://github.com/gardener/gardener-extension-shoot-dns-service/blob/v1.46.0/charts/gardener-extension-shoot-dns-service/values.yaml#L52 it should be:
```suggestion
    imageVectorOverwrite: |
```

robertvolkmann (changes requested):

robertvolkmann:
The overwriting the image vector has not the desired effect, because the image of the `dns-controller-manager` can only be configured by the removed value, see 
https://github.com/gardener/gardener-extension-shoot-dns-service/blob/release-v1.47/charts/gardener-extension-shoot-dns-service/templates/dnsman-deployment.yaml#L50

control-plane/roles/gardener/templates/shoot-dns-service/controller-deployment.yaml
Gerrit91: I cannot test it. If you are sure it works, we can do it.

Gerrit91: Added a test now. You missed a parenthesis but apart from this it worked fine. :)

robertvolkmann (approved):

PR #338, opened 2024-10-23
Fix shoot-dns-service deployment.

Gerrit91:
## Description

None


Gerrit91:
Was a mistake, this happens with shoot-dns-extension v1.49.

PR #345, opened 2024-10-31
Configure VRFs mentioned in sonic_ports_dict, sonic_vlans

### LINKED ISSUE #343: Add option to configure VRFs in SONiC

mwindower:
## Description

This eases the work with the sonic role because we can configure VRFs also for ports that do not speak BGP: plain routed or plain VLAN ports.

References:
Closes #343 

/cc @iljarotar @majst01


iljarotar (approved):

robertvolkmann (changes requested):
Can we move the additional logic into Ansible task, something like:
```
    - name: Set facts in a loop
      set_fact:
        my_list: "{{ my_list | default([]) + [ item ] }}"
      loop:
        - apple
        - banana
        - cherry
```

robertvolkmann (approved):

PR #348, opened 2024-11-07
Check if INTERFACE needs to be added to config_db in SONiC role

iljarotar:
The template for the config_db.json currently has the problem of potentially adding `INTERFACE: null` to the config_db, which ruins all interface configuration. Unfortunately the scoping of jinja2 variables makes it impossible to change a variable value inside of a scope and access the updated value from a different scope (See [docs](https://jinja.palletsprojects.com/en/stable/templates/#assignments)). So I ended up using these complicated chains of filters. Better ideas are welcome.


robertvolkmann:
@iljarotar please add a test for this case.

PR #349, opened 2024-11-07
Fix deployment of updated networking-cilium extension.

Gerrit91:
## Description

References https://github.com/metal-stack/releases/pull/200.


robertvolkmann (approved):
Good catch.

PR #352, opened 2024-11-14
Create a bgp-neighbor stats and mount result to metal-core

majst01:
## Description

Related to:

- [ ] https://github.com/metal-stack/metal-api/issues/590


partition/roles/metal-core/files/bgp-neighbor-stats.service
```diff
@@ -0,0 +1,6 @@
+[Unit]
+Description=Docker System Prune
```
robertvolkmann, line 2: Wrong description.

partition/roles/metal-core/files/bgp-neighbor-stats.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -e
```
robertvolkmann, line 3:
```suggestion
set -o errexit
```

partition/roles/metal-core/files/bgp-neighbor-stats.service
```diff
@@ -0,0 +1,6 @@
...
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/bgp-neighbor-stats.sh
```
robertvolkmann, line 6: Is it necessary to run as root user?

partition/roles/metal-core/tasks/main.yaml
```diff
@@ -46,6 +46,41 @@
     - frr-reload.service
   when: metal_stack_switch_os_is_sonic
 
+- name: create bgp-neighbor-stats directory
+  file:
+    path: /var/run/bgp-neighbors
+    state: directory
```
robertvolkmann, line 52: owner, group and mode are missing.

partition/roles/metal-core/tasks/main.yaml
```diff
@@ -46,6 +46,41 @@
...
+    group: root
+    mode: "0755"
+
+- name: create bgp-neighbor-stats.service
+  copy:
+    src: bgp-neighbor-stats.service
+    dest: /etc/systemd/system/bgp-neighbor-stats.service
+    owner: root
+    group: root
+    mode: "0644"
```
robertvolkmann, line 68: daemon-reload is necessary!

partition/roles/metal-core/files/bgp-neighbor-stats.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -e 
+tmpfile=$(mktemp /var/run/bgp-neighbors/bgp-neighbors.XXXXXX)
+destfile=/var/run/bgp-neighbors/bgp-neighbors.json
+
+/usr/bin/vtysh -c "show ip bgp vrf all neighbors json" > "${tmpfile}"
+
+mv "${tmpfile}" "${destfile}"
+rm "${tmpfile}"
```
robertvolkmann, line 10: If the script fails for any reason, `metal-core` still believes there are BGP neighbors. Please remove the destination file at the beginning of the script.

robertvolkmann (changes requested):
Please move the bgp neighbor stats stuff in a new task and make it optional.

partition/roles/metal-core/files/bgp-neighbor-stats.service
majst01: Yes, because it must be compatible with cumulus and sonic, where we have two different main users, (cumulus and admin) and i dont want to create a new user for this purpose.

majst01:
> Please move the bgp neighbor stats stuff in a new task and make it optional.

why optional, metal-core will depend on this after the implementation of this feature is finished.

robertvolkmann:
> > Please move the bgp neighbor stats stuff in a new task and make it optional.
> 
> why optional, metal-core will depend on this after the implementation of this feature is finished.

What is the expected behavior of this feature? What will happen if no BGP state is reported for a port?

partition/roles/metal-core/files/bgp-neighbor-stats.service
robertvolkmann: Good point.

majst01:
> Does that mean I can no longer replace a broken switch with a new one?

You can, its all about visibility in the first place.

robertvolkmann:
So no one is really depending on this feature, and metal-core should be able to function without it. Could you consider making it optional in case we need to disable it in production for any unforeseen reason?

majst01:
> So no one is really depending on this feature, and metal-core should be able to function without it. Could you consider making it optional in case we need to disable it in production for any unforeseen reason?

No, metal-core depends on it with: https://github.com/metal-stack/metal-core/pull/136

Making it optional means more code in metal-core and metalctl, this is a highly useful feature for a stable dataplane and i cannot see any reason why it should not be used.

robertvolkmann:
It is already optional in metal-core. If `bgpNeighborStateFile` is empty, no new functionality is called.

partition/roles/metal-core/files/bgp-neighbor-stats.timer
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Collect BGP Neighbor statistics
+
+[Timer]
+OnCalendar=*:0/2
```
robertvolkmann, line 5: Is it triggered only every two minutes to reduce the load on the system?

partition/roles/metal-core/files/bgp-neighbor-stats.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -o errexit
+
+tmpfile=$(mktemp /var/run/bgp-neighbors/bgp-neighbors.XXXXXX)
+destfile=/var/run/bgp-neighbors/bgp-neighbors.json
+
+/usr/bin/vtysh -c "show ip bgp vrf all neighbors json" > "${tmpfile}"
+
+rm -f "${destfile}"
```
robertvolkmann, line 10:
```suggestion
rm -f "${destfile}"

tmpfile=$(mktemp /var/run/bgp-neighbors/bgp-neighbors.XXXXXX)
destfile=/var/run/bgp-neighbors/bgp-neighbors.json

/usr/bin/vtysh -c "show ip bgp vrf all neighbors json" > "${tmpfile}"
```
`mktemp` or `vtysh` may fail.

majst01: What exactly are you suggesting here, don't get it ?

robertvolkmann: The same reason as above:
> If the script fails for any reason, metal-core still believes there are BGP neighbors. Please remove the destination file at the **beginning** of the script.

majst01: But if anything fails, the mv will not happen, also metal-core checks for mtime <= 1 hour

robertvolkmann: Thank you, I didn't know about that.

robertvolkmann (approved):

mwennrich (approved):

PR #358, opened 2024-11-21
Allow individual config_db per host via ztp.json

iljarotar:
## Description

Initially, this PR was intended to enable a ZTP-only deployment for SONiC switches, such that all configuration currently handled by the `sonic` role could be managed entirely by ZTP. The motivation behind a ZTP-only approach is that the template for the `config_db.json` has become quite complicated and difficult to debug recently, because many special cases needed to be handled. With a `ztp.json` file it is possible to provide an individual `config_db.json` for each switch, which makes templating obsolete.

The downside to this way of provisioning is that the SONiC config cannot be changed via CI anymore, but only by resetting the switch completely. For the `config_db.json`, this seems fine to me, as it reduces the risk of breaking the switch by providing an invalid breakout config for example. However, providing the `frr.conf` file via ZTP doesn't seem a good idea to me and it would have introduced a lot of redundancy, because the `sonic` role also renders the `frr.conf`.

Therefore, I decided to make `ztp.json` only provide the `config_db.json` and some setup scripts, but leave the rest to the `sonic` role. To avoid conflicts with the `config_db.json` rendered by the `sonic` role, I introduced a new variable `sonic_render_config_db_template` which defaults to `true`, to not break existing deployments.

### Noteworthy

```NOTEWORTHY
It is now possible to provision SONiC switches with a static `config_db.json` provided through ZTP instead of rendering this configuration dynamically during deployment. This approach leads to more stable rollouts and might become our default for SONiC deployments. When using this new provisioning variant, follow the example in the `ztp` role documentation and set `sonic_render_config_db_template` in the `sonic` role to false. Also leave `sonic_ports`, `sonic_breakouts` and `sonic_portchannels` empty. Otherwise there might be conflicts between the provided `config_db.json` and the one rendered by the `sonic` role.
```


partition/roles/ztp/README.md
```diff
@@ -4,16 +4,19 @@ Configures a server for providing zero-touch-provisioning scripts for switches.
...
+| ztp_port                     |           | the port to serve ztp scripts on.                                                                         |
+| ztp_authorized_keys          | yes       | the authorized keys that should be installed by ztp.                                                      |
+| ztp_admin_user               |           | the user for which the authorized keys will be provisioned.                                               |
+| ztp_additional_files         |           | puts additional files into serve directory.                                                               |
+| ztp_sonic_nameservers        |           | the nameservers to put into resolv.conf for sonic                                                         |
+| ztp_sonic_extended_cacl.ipv4 |           | iptables ipv4 rules that should be added as extended Control Plane ACLs (Edgecore Sonic specific feature) |
+| ztp_sonic_extended_cacl.ipv6 |           | iptables ipv6 rules that should be added as extended Control Plane ACLs (Edgecore Sonic specific feature) |
```
Gerrit91, line 19: You need to explain why SONiC variables appear in this role in a paragraph / section at the top of the role. So what the intended purpose is and how it is supposed to be used.

Probably it's also worth to add a `NOTEWORTHY` section if this PR means that the way of how SONiC switches are supposed to be bootstrapped change.

robertvolkmann:
The PR should ensure that AgentX is enabled on the leaves.

robertvolkmann (changes requested):
The PR should ensure that AgentX is enabled on the leaves.

Gerrit91:
@mreiger yesterday brought up that it might be good to add a comment to the documentation that with this approach it is not possible to simply add something new to the `config_db.json`. Every change there requires re-provisioning the switch.

partition/roles/ztp/templates/ztp.json.j2
```diff
@@ -5,17 +5,35 @@
...
       "plugin": {
-        "url": "http://{{ ztp_listen_address }}:{{ ztp_port }}/reload.sh"
-      }
+        "url": "http://{{ ztp_listen_address }}:{{ ztp_port }}/ztp-provisioning-script.sh"
+      },
+      "shell": true,
+      "reboot-on-success": true
     },
+{% endif %}
     "restart-ztp-no-config": false
```
robertvolkmann, line 37: Why do we need this?

partition/roles/ztp/templates/ztp.json.j2
```diff
@@ -5,17 +5,35 @@
...
-    "04-reload": {
+{% if ztp_provisioning_script %}
+    "05-provisioning-script": {
       "plugin": {
-        "url": "http://{{ ztp_listen_address }}:{{ ztp_port }}/reload.sh"
-      }
+        "url": "http://{{ ztp_listen_address }}:{{ ztp_port }}/ztp-provisioning-script.sh"
+      },
+      "shell": true,
+      "reboot-on-success": true
```
robertvolkmann, line 34: Should it be elevated by one level, so that the switch will reboot even if no provisioning script has been set up?

partition/roles/ztp/templates/ztp.json.j2
```diff
@@ -5,17 +5,35 @@
...
     },
-    "04-reload": {
+{% if ztp_provisioning_script %}
+    "05-provisioning-script": {
       "plugin": {
-        "url": "http://{{ ztp_listen_address }}:{{ ztp_port }}/reload.sh"
-      }
+        "url": "http://{{ ztp_listen_address }}:{{ ztp_port }}/ztp-provisioning-script.sh"
+      },
+      "shell": true,
```
robertvolkmann, line 33: Why do we need `shell: true` here but not for `ztp.sh`?

partition/roles/ztp/templates/ztp.json.j2
```diff
@@ -5,17 +5,35 @@
         "url": "http://{{ ztp_listen_address }}:{{ ztp_port }}/ztp.sh"
```
robertvolkmann, line 5: What do you think about remaining it to `user.sh`?
```suggestion
        "url": "http://{{ ztp_listen_address }}:{{ ztp_port }}/user.sh"
```

partition/roles/ztp/templates/ztp.json.j2
iljarotar: Don't know why it's needed, but it didn't work without it.

partition/roles/ztp/templates/ztp.json.j2
iljarotar: Done

partition/roles/ztp/templates/ztp.json.j2
iljarotar: Done

partition/roles/ztp/templates/ztp.json.j2
iljarotar: Removed `"restart-ztp-no-config": false`, it was a remainder from the previous approach.

robertvolkmann (approved):

partition/roles/sonic/tasks/main.yaml
```diff
@@ -58,62 +36,9 @@
...
-  set_fact:
-    config_db: "{{ lookup('template', 'metal.yaml.j2') }}"
-
-- name: Save config_db as JSON file
-  copy:
-    content: "{{ config_db | from_yaml | to_nice_json }}"
-    dest: /etc/sonic/config_db.json
-  notify: "config {{ sonic_config_action }}"
+- name: Render and save config_db
+  include_tasks: config_db.yaml
```
robertvolkmann, line 40:
```suggestion
  import_tasks: config_db.yaml
```

robertvolkmann (changes requested):

robertvolkmann (approved):

Issue #357, opened 2024-11-21
Fix order of dhcp tasks


iljarotar:
The dhcp role installs isc-dhcp-server first and renders the config after that. This causes isc-dhcp-server to crash with error `Not configured to listen on any interfaces!`. Perhaps the config should be rendered before installing.

PR #359, opened 2024-11-21
Install isc-dhcp-server after rendering configs

iljarotar:
Closes https://github.com/metal-stack/metal-roles/issues/357


robertvolkmann:
Do you understand why we're not notify a handler to restart `isc-dhcp-server` when its configuration changed?
```
tasks:
- name: render dhcpd conf
  template:
    src: dhcpd.conf.j2
    dest: /etc/dhcp/dhcpd.conf
  notify: Restart isc-dhcp-server
  - …

handlers:
  - name: Restart isc-dhcp-server
    service:
      name: "{{ dhcp_service_name }}"
      state: restarted
```

I also miss a `daemon-reload: true`.

Gerrit91 (approved):

PR #368, opened 2025-01-10
wait for connection after reboot

iljarotar:
## Description

This is an attempt to fix the issue, where a server gets rebooted and the connection times out despite the server having successfully booted.
As described [here](https://github.com/ansible/ansible/issues/79992), the `reboot` module sometimes has issues with reconnecting, while using `wait_for_connection` after `reboot` seems to work more reliably.


iljarotar:
Turns out, the issue was caused by selecting the wrong runner. Closing.

PR #372, opened 2025-02-06
Allow configuring static ports in SONiC role for machines

iljarotar:
## Description

This PR allows static port configuration on SONiC switches (including leaf switches). This comes in handy if there are machines connected to the fabric that are not managed by the metal-stack. So configuration by metal-core and the registration process of the metal-api is circumvented.


Gerrit91 (dismissed):

iljarotar:
@mreiger could you open a new PR for your changes?

partition/roles/ztp/defaults/main/main.yaml
```diff
@@ -11,3 +11,5 @@ ztp_provisioning_script:
 ztp_additional_files: []
 # - name: foo.sh
 #   data: echo
+
+ztp_provisioning_script:
```
Gerrit91, line 15: Can be removed, it's already defined in the file.

Gerrit91:
As this PR was originally for a different purpose and @iljarotar already moved the changes related to his concerns to another PR, I would like to propose to re-open a new PR for this in order to align the reviewers and assignees. Hope this is not too much of work but it should be fairly easy to do this.

PR #374, opened 2025-02-10
Support gardener-extension-shoot-cert-service v1.45.0

robertvolkmann:
## Description

None


Gerrit91 (commented):
API version also has to change from `v1beta1`:

```yaml
apiVersion: core.gardener.cloud/v1
```

Gerrit91 (approved):

PR #375, opened 2025-02-10
Allow shooted seeds to be deployed with HA control plane.

Gerrit91:
## Description

See description for the release notes.

### Noteworthy

```NOTEWORTHY
As with g/g v1.105 the `VPAForETCD` feature gate gets enabled unconditionally it is advised to use high-availiability for shoots to prevent unavailability of the ETCD during scaling. For the shooted seeds this can be configured by setting the new field `gardener_shooted_seed.high_availability_control_plane` to either `node` or `zone`.
```

### DIFF
--- a/control-plane/roles/gardener/defaults/main/gardener.yaml
+++ b/control-plane/roles/gardener/defaults/main/gardener.yaml
@@ -101,6 +101,8 @@ gardener_shooted_seeds: []
 #   verticalPodAutoscaler: false
 #   pod_cidr:
 #   service_cidr:
+#   # possible values are zone and node, please refer to the gardener documentation to what is appropriate for your environment:
+#   # https://gardener.cloud/docs/guides/high-availability/control-plane/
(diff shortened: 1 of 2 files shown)


robertvolkmann (approved):

control-plane/roles/gardener/defaults/main/gardener.yaml
```diff
@@ -101,6 +101,7 @@ gardener_shooted_seeds: []
 #   verticalPodAutoscaler: false
 #   pod_cidr:
 #   service_cidr:
+#   high_availability_control_plane: node
```
majst01, line 104: Maybe nice to have some explanation what types are possible (node|zone)

mwennrich (approved):

control-plane/roles/gardener/defaults/main/gardener.yaml
Gerrit91: Done.

PR #376, opened 2025-02-10
Use etcd-backup-restore images from ghcr.io.

Gerrit91:
## Description

As unfortunately we are far behind the upstream etcd-backup-restore of the virtual garden we are running into the issue that we still use `eu.gcr.io`, which is shutting down on 28th of February. 

It is hard to update to ETCD deployment because it had breaking changes in the past. One attempt how it could be done I made [here](https://github.com/metal-stack/metal-roles/commit/e7578e19ac4bee2bd5e8a6900ae26bcb5a0828ac), but with the migration to the [gardener operator](https://github.com/metal-stack/metal-roles/pull/360) this would be quite unnecessary because the Gardener Operator manages the virtual garden ETCD on its own.

Meanwhile, this PR enables operators to overwrite the ETCD image to point to another mirror `ghcr.io/metal-stack/gardener/etcd` and `ghcr.io/metal-stack/gardener/etcdbrctl`.

### Required Actions

```ACTIONS_REQUIRED
As `eu.gcr.io` is shutting down on 28th of February we offer operators the possibility to point the virtual garden ETCD images to `ghcr.io/metal-stack`. This will be done automatically through the release vector and will be removed as soon as migrated to the Gardener Operator.
```


majst01 (approved):

robertvolkmann:
@Gerrit91, this must be incorporated into the release vector, as the release will be at risk starting in March without it.

Gerrit91:
@robertvolkmann It's now part of the release vector.

robertvolkmann (approved):

PR #379, opened 2025-02-19
Stretched exits for sonic

Honigeintopf:
## Description

In order to configure multiple neighbors for the exits, which was done before only with cumulus, this pr introduces functionality to do so.
Tests for this were also written, this is backwards compatible.


Honigeintopf:
No will close this PR. Is the exact same as #421

PR #380, opened 2025-02-19
feat: allow configuring gardeners vali

vknabel:
## Description

In some cases we need the logging stack but without vali. This is now possible by setting `gardener_logging_vali_enabled: false`. Default conforms with gardeners default of true.

Needs to be tested before merge.


robertvolkmann:
Why about using `gardener_logging_vali_enabled`?

control-plane/roles/gardener/templates/gardenlet-values.j2
```diff
@@ -25,6 +25,8 @@ config:
 {% if gardener_logging_enabled %}
   logging:
     enabled: true
+    vali:
+      enabled: {{ gardener_logging_vali_enabled | lower  }}
```
robertvolkmann, line 29:
```suggestion
      enabled: {{ gardener_logging_vali_enabled | lower }}
```

robertvolkmann (approved):

PR #383, opened 2025-02-21
Adjustments to make static machine ports possible

mreiger:
(cherry picked from commit 0342e666150c0c005fbbd2ff79c65a5276c23119)

## Description

Adjustments to the sonic role templates to make static machine ports possible on switches without metal-core; eg for some storage systems that can speak BGP.

Re-using the `sonic_interconnects` tree for this because it already contained almost all that was necessary.


iljarotar:
Closing this as it seems to only make changes in the deprecated sonic role and the changes have already been made to the new role. Please re-open if necessary.

PR #389, opened 2025-03-06
Configure lldp hello timer on upstream sonic

iljarotar:
## Description

For edge-core SONiC we use a field in the `config_db.json` to configure LLDP hello timer. This field is missing in upstream SONiC, therefore `lldpcli` is used. Running the same command on an edge-core SONiC version 202111.4 works as well, so this change doesn't break anything.

### DIFF
--- a/partition/roles/sonic/tasks/main.yaml
+++ b/partition/roles/sonic/tasks/main.yaml
@@ -60,5 +60,10 @@
   notify:
   - restart caclmgrd
 
+# upstream sonic cannot configure lldp so we set it here via lldpcli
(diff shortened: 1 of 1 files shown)


majst01 (approved):

robertvolkmann:
This change is not permanent and will be lost if the LDAP container is restarted.

Additionally, we employ the same method in the mini-lab as well: https://github.com/metal-stack/mini-lab/blob/master/roles/sonic/tasks/main.yaml#L9

iljarotar:
@robertvolkmann is right, this change is not permanent. Will try so solve it another way.

PR #390, opened 2025-03-11
Add a new sonic-config role for SONiC configuration.

iljarotar:
## Description

This PR introduces a new sonic-config role to replace the existing sonic role. Most of the variables work the same way they do for the sonic role. But instead of using a jinja template to render a config_db.json, the [ConfigDB utils](https://github.com/metal-stack/sonic-configdb-utils) are used. It's a tool that uses mostly the same logic as the template but adds more validation and simplifies some of the logic as well. 

Furthermore, I added a static minimal config_db.json to the ztp role and removed the optional `ztp_provisioning_script` as it was only used as a workaround to replace the broken dhcp_relay, which is not necessary anymore.

Since https://github.com/metal-stack/ansible-common/pull/31 it is possible to integrate metal-core into SONiC in such a way that it gets restarted along with the other components when a `config reload` occurs. For this I added a `wantedby: sonic.target` to the metal-core deployment.

Depends on https://github.com/metal-stack/sonic-configdb-utils/pull/18.

### Required Actions

```ACTIONS_REQUIRED
Read the sonic-config README for a migration guide from the old sonic role to the new sonic-config role.
```


partition/roles/ztp/tasks/main.yaml
```diff
@@ -26,11 +26,14 @@
...
+- name: copy frr.conf and config_db.json
   copy:
-    dest: "{{ ztp_host_dir_path }}/config/frr.conf"
-    src: frr.conf
+    dest: "{{ ztp_host_dir_path }}/config/{{ item }}"
+    src: "{{ item }}"
     mode: 0644
+  loop:
+    - frr.conf
+    - config_db.json
```
Gerrit91, line 36: Before this, the role was agnostic in terms of switch operating system, right? Does it make sense to have this file Cumulus?

Gerrit91:
For the release notes: If there is a required actions section you do not necessarily need to document the breaking change as otherwise when reading the release notes the information is split across the release notes, which makes it harder to comprehend.

partition/roles/ztp/tasks/main.yaml
iljarotar: No it wasn't agnostic before. There is a `ztp.json` file and a `ztp_provisioning_script` variable which are only relevant for SONiC. Also the `frr.conf` isn't used by Cumulus deployments.

partition/roles/sonic/tasks/generate_config_db.yaml
```diff
@@ -0,0 +1,45 @@
...
+  copy:
+    src: /etc/sonic/config_db.json
+    dest: /tmp/sonic/config_db.json
+    mode: 0644
+    remote_src: true
+
+- name: Run sonic-configdb-utils
+  community.docker.docker_container:
+    name: sonic-configdb-utils
+    image: "ghcr.io/metal-stack/sonic-configdb-utils:{{ sonic_configdb_utils_image_tag }}"
```
Gerrit91, line 22: Should be moved to into the release vector. I can help you with that.

Gerrit91 (commented):
I think the release notes saying that people can use `""` as a value are not correct anymore.

partition/roles/sonic/tasks/generate_config_db.yaml
```diff
@@ -0,0 +1,45 @@
...
+    pull: true
+    detach: false
+    cleanup: true
+    command: generate
+    volumes:
+      - /usr/share/sonic/device:/usr/share/sonic/device:ro
+      - /tmp/sonic:/etc/sonic:rw
+      - /tmp/sonic:/sonic:ro
+
+- name: Copy generated config_db.json to /etc/sonic
```
robertvolkmann, line 32:
```suggestion
- name: Overwrite the existing config_db.json with the newly generated one
```
No backup of the existing one?

partition/roles/sonic/tasks/generate_config_db.yaml
```diff
@@ -0,0 +1,45 @@
...
+  template:
+    src: sonic-config.yaml.j2
+    dest: /tmp/sonic/sonic-config.yaml
+
+- name: Copy current config_db.json into temporary directory
+  copy:
+    src: /etc/sonic/config_db.json
+    dest: /tmp/sonic/config_db.json
+    mode: 0644
+    remote_src: true
```
robertvolkmann, line 17: What happens if no `config_db.json` already exists?

partition/roles/sonic/tasks/generate_config_db.yaml
```diff
@@ -0,0 +1,45 @@
...
+- name: Run sonic-configdb-utils
+  community.docker.docker_container:
+    name: sonic-configdb-utils
+    image: "{{ sonic_configdb_utils_image_name }}:{{ sonic_configdb_utils_image_tag }}"
+    pull: true
+    detach: false
+    cleanup: true
+    command: generate
+    volumes:
+      - /usr/share/sonic/device:/usr/share/sonic/device:ro
```
robertvolkmann, line 28: Why you mount the whole device tree instead of only the directories `hwsku` and `platform`?
```
$ ls -l /usr/share/sonic/
total 16
drwxr-xr-x 1 root root 4096 Feb 13 10:11 device
lrwxrwxrwx 1 root root   69 Feb 13 10:11 hwsku -> /usr/share/sonic/device/x86_64-accton_as7726_32x-r0/Accton-AS7726-32X
lrwxrwxrwx 1 root root   51 Feb 13 10:11 platform -> /usr/share/sonic/device/x86_64-accton_as7726_32x-r0
drwxr-xr-x 2 root root 4096 Feb 13 10:11 scripts
drwxr-xr-x 2 root root  710 Jun 19  2024 templates
```

partition/roles/sonic/tasks/generate_config_db.yaml
```diff
@@ -0,0 +1,45 @@
...
+  copy:
+    src: /etc/sonic/config_db.json
+    dest: /tmp/sonic/config_db.json
+    mode: 0644
+    remote_src: true
+
+- name: Run sonic-configdb-utils
+  community.docker.docker_container:
+    name: sonic-configdb-utils
+    image: "{{ sonic_configdb_utils_image_name }}:{{ sonic_configdb_utils_image_tag }}"
```
robertvolkmann, line 22: Why running the `sonic-configdb-utils` on the host instead of running it in the `metal-stack-deployment` container?

partition/roles/ztp/README.md
```diff
@@ -22,50 +22,15 @@ host leaf01 {
...
-    data: ...
-```
-
-When a SONiC switch is deployed via `ztp.json` and configured by the `sonic` role afterwards, make sure to leave the `sonic_ports`, `sonic_portchannels` and `sonic_breakouts` variables empty and set `sonic_render_config_db_template` to false.
-Otherwise the `sonic` role will override the `config_db.json` provided by the `ztp.json`.
-The result of this may not be intended and, in the worst case, the switch will reach a broken state from which it only can be restored by a factory reset.
-Of course it is also possible to load only a minimal `config_db.json` via ZTP and allow the `sonic` role to render its template based on the `sonic_ports`, `sonic_portchannels` and `sonic_breakouts` variables.
-Both approaches have their pros and cons.
-
-### Pros and Cons of Loading a Static config_db.json via ZTP
```
robertvolkmann, line 43: Why you stop using this approach?

partition/roles/ztp/files/ztp-provisioning-script.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+sudo config-setup factory
```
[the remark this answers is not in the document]
iljarotar: Creates a default config_db.json to ensure there is one (see below)

partition/roles/sonic/tasks/generate_config_db.yaml
iljarotar: No backup. We never made backups of the existing ones.

partition/roles/sonic/tasks/generate_config_db.yaml
iljarotar: See above. `sudo config-setup factory` ensures there is one.

partition/roles/sonic/tasks/generate_config_db.yaml
iljarotar: Depending on the SONiC version the `platform` directory might be empty.

partition/roles/sonic/tasks/generate_config_db.yaml
iljarotar: It needs to read the `/etc/sonic` and `/usr/share/sonic/device` directories. The required files could be copied to the control node so `sonic-confidb-utils` could be run there, but what for? What is the benefit of running it inside the `metal-stack-deployment` container?

partition/roles/ztp/README.md
iljarotar: As @Gerrit91 once pointed out to me, using an approach where large parts of the configuration are applied only once during initial setup and never touched afterwards can add to a certain fear of 'pressing the button' when something needs to be changed. Therefore I decided to go back to the approach where the config is generated on every run of the sonic role.

partition/roles/ztp/files/ztp-provisioning-script.sh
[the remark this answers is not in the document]
robertvolkmann: This may introduce an unwanted default configuration. Depending on the SONiC version and the platform a "wrong"  preset could be configured in `default_sku`.

partition/roles/sonic/tasks/generate_config_db.yaml
```diff
@@ -0,0 +1,45 @@
...
+    pull: true
+    detach: false
+    cleanup: true
+    command: generate
+    volumes:
+      - /usr/share/sonic/device:/usr/share/sonic/device:ro
+      - /tmp/sonic:/etc/sonic:rw
+      - /tmp/sonic:/sonic:ro
```
robertvolkmann, line 30: Why do you need the same directory at two different locations?

partition/roles/ztp/files/ztp-provisioning-script.sh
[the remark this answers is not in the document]
iljarotar: Could you give me an example?

partition/roles/ztp/files/config_db.json
```diff
@@ -0,0 +1,7 @@
+{
+  "DEVICE_METADATA": {
+    "localhost": {
+      "docker_routing_config_mode": "split-unified"
```
iljarotar, line 4: TODO: on EC SONiC this is called `split`, for upstream we need `split-unified`. Maybe introduce dynamic config_db.json depending on the SONiC version.

iljarotar: Adapted image instead.

iljarotar:
Works in the test environment.

partition/roles/sonic/templates/sonic-config.yaml.j2
```diff
@@ -0,0 +1,31 @@
...
+features: {{ sonic_features }}
+frr_mgmt_framework_config: {{ sonic_frr_mgmt_framework_config }}
+interconnects: {{ sonic_interconnects }}
+hostname: {{ sonic_hostname }}
+lldp_hello_time: {{ sonic_lldp_hello_timer }}
+loopback_address: {{ sonic_loopback_address }}
+mclag: {{ sonic_mclag }}
+mgmt_interface:
+  gateway_address: {{ sonic_mgmtif_gateway }}
+  ip: {{ sonic_mgmtif_ip }}
```
robertvolkmann, line 13: How do you ensure that `sonic_mgmtif_gateway` and `sonic_mgmtif_ip` are defined?

partition/roles/sonic/templates/sonic-config.yaml.j2
```diff
@@ -0,0 +1,31 @@
...
+ntp:
+  src_interface: {{ sonic_ntp.src_interface | default('') }}
+  servers: {{ sonic_ntpservers }}
+  vrf: {{ sonic_ntp.vrf | default('') }}
+portchannels:
+  default_mtu: {{ sonic_portchannels_default_mtu }}
+  list: {{ sonic_portchannels }}
+ports:
+  default_fec: {{ sonic_ports_default_fec }}
+  default_mtu: {{ sonic_ports_default_mtu }}
```
robertvolkmann, line 25: How do you ensure that `sonic_ports_default_mtu` is defined?

partition/roles/sonic/templates/sonic-config.yaml.j2
```diff
@@ -0,0 +1,31 @@
...
+  list: {{ sonic_ports }}
+sag: {{ sonic_sag }}
+ssh_sourceranges: {{ sonic_ssh_sourceranges }}
```
robertvolkmann, line 28: How do you ensure that `sonic_ssh_sourceranges` is defined?

robertvolkmann (changes requested):

robertvolkmann (approved):
@Gerrit91 any comments about handling the default values?

partition/roles/sonic-config/README.md
```diff
@@ -0,0 +1,499 @@
...
+From version 202111.6 to version 202111.7 some change in the dhcp_relay broke the relay's ability to forward PXE boot OFFER messages from the server back to the client.
+This bug was fixed in version 202111.11.
+
+## Migration from the Deprecated sonic Role
+
+Most of the variables have the same name, meaning and structure as the correspondig ones in the sonic role.
+So in most cases simply adding `_config` to the variable name is enough, e.g. `sonic_interconnects` becomes `sonic_config_interconnects`.
+Variables that were changed, removed or added are explained below.
+
+### sonic_ntpservers
```
Gerrit91, line 28: These headings I think should be wrapped in inline code because it somehow is inline-code.

```suggestion
### `sonic_ntpservers`
```

partition/roles/sonic-config/README.md
```diff
@@ -0,0 +1,499 @@
...
+# Image of the [config generator tool](https://github.com/metal-stack/sonic-configdb-utils).
+sonic_configdb_utils_image_name: # ghcr.io/metal-stack/sonic-configdb-utils
+
+# Image tag (should be `ec202111` for Edgecore SONiC versions 202111.x).
+sonic_configdb_utils_image_tag: # ec202111
+
+# How to handle routing configuration.
+# With split mode all routing configuration is delegated to FRR.
+# With unified mode the FRR config gets generated by SONiC from its ConfigDB.
+sonic_config_docker_routing_config_mode: # split
```
Gerrit91, line 189: It would be nice to see possible values for these kind of enums. From the starting text I assume that the second mode is `split-unified`?

partition/roles/sonic-config/README.md
```diff
@@ -0,0 +1,513 @@
...
+    speed: 40000
+    vrf: Vrf45
+```
+
+to
+
+```yaml
+sonic_config_ports:
+  # New field to configure default value for auto negotiation.
+  default_autoneg: # on
```
robertvolkmann, line 105: Why on is commented out?

partition/roles/sonic-config/README.md
```diff
@@ -0,0 +1,513 @@
...
+  default_fec: none
+  default_mtu: 9000
+  list:
+    # New field to configure auto negotiation per port.
+    - autoneg: # off
```
robertvolkmann, line 110: Why off is commented out?

partition/roles/sonic-config/README.md
```diff
@@ -0,0 +1,513 @@
...
+# - separated: FRR config generated from ConfigDB, each FRR daemon has its own config file
+# - unified: FRR config generated from ConfigDB, single FRR config file
+# - split: FRR config not generated from ConfigDB, each FRR daemon has its own config file
+# - split-unified: FRR config not generated from ConfigDB, single FRR config file"
+#
+# Unfortunately, Edgecore SONiC versions 202111.x handle these values slightly differently and
+# documentation is hard to come by. Most importantly, `split-unified` does not exist on versions
+# 202111.x. Use `split` instead. When metal-core is deployed on the switch, `split` is required
+# as metal-core will write a frr.conf, which will not be persisted otherwise.
+sonic_config_docker_routing_config_mode: # split
```
robertvolkmann, line 203: Why split it commented out?

partition/roles/sonic-config/defaults/main/main.yaml
```diff
@@ -0,0 +1,30 @@
+sonic_config_bgp_ports: []
+sonic_config_breakouts: {}
+sonic_configdb_utils_image_tag: ec202111
```
robertvolkmann, line 3: Image tag should come from release vector.

partition/roles/sonic-config/README.md
iljarotar: All example values are commented out (with a few exceptions). I wanted to make a clear distinction between the variables that are part of the role and their example values. Only in such cases like 

```yaml
# Features to be monitored by the container_checker.
sonic_config_features:
  # Name of the container.
  some-feature:
    # If enabled the container will get restarted on failure.
    auto_restart: # true
    # If enabled the container_checker will monitor the container.
    enabled: # true
```

it would have looked weird to comment `some-feature` out.

partition/roles/sonic-config/README.md
iljarotar: Auto negotiation config is a new feature of this role which doesn't exist in the `sonic` role. The migration guide shows what changes need to be made to adapt to the new role. Since auto negotiation wasn't configured before it is documented here but with a comment to avoid the misunderstanding that `default_autoneg: on` is required when migrating to the new role.

iljarotar: EDIT: removed all comments

PR #392, opened 2025-03-12
stop mirroring from eu.gcr.io

mwennrich:
## Description

```
"error":"GET https://eu.gcr.io/v2/gardener-project/hyperkube/tags/list?n=1000: DENIED: Failed to read tags for host 'eu.gcr.io'
```
=> eu.gcr.io has been disabled, stop trying to mirror from it

### DIFF
--- a/control-plane/roles/isolated-clusters/defaults/main/images.yaml
+++ b/control-plane/roles/isolated-clusters/defaults/main/images.yaml
@@ -21,10 +21,6 @@ isolated_clusters_registry_oci_mirror_config:
       destination: http://registry:5000/lightos-csi/lb-nvme-discovery-client
       match:
         semver: ">= 1.14.0"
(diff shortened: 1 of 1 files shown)


control-plane/roles/isolated-clusters/defaults/main/images.yaml
```diff
@@ -21,10 +21,6 @@ isolated_clusters_registry_oci_mirror_config:
       destination: http://registry:5000/lightos-csi/lb-nvme-discovery-client
       match:
         semver: ">= 1.14.0"
-    - source: eu.gcr.io/gardener-project/hyperkube
```
majst01, line 24: but we need to add the new source

majst01 (approved):

majst01 (approved):

Gerrit91 (approved):

PR #401, opened 2025-03-26
Add retries to wait for metal-core to come up

iljarotar:
## Description

Sometimes when metal-core does not come up a simple restart can fix it. Therefore, instead of simply waiting for metal-core to come up, a restart will be attempted 5 times before the task fails.

### DIFF
--- a/partition/roles/metal-core/tasks/main.yaml
+++ b/partition/roles/metal-core/tasks/main.yaml
@@ -116,4 +116,4 @@
     retry_counter: 0
   include_tasks: wait_for_metal_core.yaml
   with_sequence: start=0 end={{ max_retries }}
-  when: "{{ result }} is defined and {{ result }} is failed"
+  until: result is defined and result is success
--- a/partition/roles/metal-core/tasks/main.yaml
+++ b/partition/roles/metal-core/tasks/main.yaml
@@ -117,4 +117,5 @@
   include_tasks: wait_for_metal_core.yaml
   with_sequence: start=0 end={{ max_retries }}
   loop_control:
-    break_when: result is success
+    break_when:
+      - result is success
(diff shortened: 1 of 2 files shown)


robertvolkmann (changes requested):
Did you find the cause of the problem and try to fix it before using a workaround?

iljarotar:
The problem is that after applying a breakout config to SONiC the ports sometimes take a while before they are ready. In certain cases the metal-core will fail to register if the breakouts in the running config are not what it was expecting. One possible solution is to wait for the breakouts to settle before finishing the sonic role. The only way I can think of to achieve this is by parsing the running config and comparing it to the expected breakout config. This seems pretty verbose and I'm not sure if it is worth it. I don't see a problem with simply restarting the metal-core a few times before failing. It does not increase the overall timeout and doesn't break anything.

iljarotar:
It's not a line in the metal-core but metal-api. https://github.com/metal-stack/metal-api/blob/3d853dbab796fa4ca96e9d10d795e9b9cf378c93/cmd/metal-api/internal/service/switch-service.go#L981

This occurs if the switch was present in the database before and had machines connected to ports with a breakout config that differs from the default. When the switch is newly set up and the breakout config is reset, then it needs to be restored before the metal-core attempts to register.

robertvolkmann:
Your method will not stop a switch from registering with an incorrect set of interfaces if those interfaces were not previously in the database.

The underlying issue is that it takes too long for the breakout to display the new Linux interfaces.

Perhaps we should consider reading the ports from the ConfigDB instead of querying the Linux interfaces: https://github.com/metal-stack/metal-core/blob/master/cmd/internal/switcher/sonic/sonic.go#L128.

iljarotar:
Good idea. Will close this one and open an issue in metal-core.

PR #406, opened 2025-04-07
Adaptions for g/g v1.111.

Gerrit91:
## Description

We forgot to add this to the metal-stack v0.21.2 release. :(


mwennrich (approved):

Gerrit91:
Maybe we are lucky because in the next version the old annotation gets deprecated, so it will be updated during the same rollout.

PR #408, opened 2025-04-08
Move encryption-key to secret

ostempel:
## Description

Moved encryptionKey of `backup-restore-sidecar` from config to secret.


control-plane/roles/rethinkdb-backup-restore/templates/rethinkdb.yaml
```diff
@@ -230,6 +234,9 @@ metadata:
 type: Opaque
 data:
   rethinkdb-password: "{{ rethinkdb_password | b64encode }}"
+{% if rethinkdb_backup_restore_sidecar_encryption_key %}
+  encryptionKey: {{ rethinkdb_backup_restore_sidecar_encryption_key | b64encode }}
+{% endif %}
```
robertvolkmann, line 239: Mixed style.

robertvolkmann (changes requested):

Gerrit91 (approved):

control-plane/roles/rethinkdb-backup-restore/templates/rethinkdb.yaml
Gerrit91: That's right, currently we have dash-case. Maybe it's more uniform to continue with it. Apart from this seems like it works. :)

ostempel: Here the encryption-key was in the wrong secret. Just fixed it.

PR #412, opened 2025-04-11
Add option to configure `imagePullSecrets` when deploying valkey

### LINKED ISSUE #411: Support passing image pull secrets for valkey deployment
https://github.com/bitnami/charts/blob/main/bitnami/valkey/values.yaml#L22

Otherwise we run into docker rate limits in the release integration.

mac641:
## Description

Closes #411.


control-plane/roles/valkey/templates/valkey-secret.yaml
```diff
@@ -0,0 +1,12 @@
+{% if valkey_registry_auth_enabled %}
```
Gerrit91, line 1: The Ansible task should have this condition in a `when` otherwise an empty yaml file would be applied.

Gerrit91 (approved):

PR #414, opened 2025-04-15
Static machine ports

mreiger:
## Description

Small changes to the sonic role so that static machine ports can be configured for machines that are not deployed through metal-stack


partition/roles/sonic/templates/frr.conf.j2
```diff
@@ -136,7 +145,9 @@ router bgp {{ sonic_asn }} vrf {{ i.vrf }}
   {{ a }}
     {% endfor %}
     {% endif %}
+    {% if i.keep_private_as is not defined or i.keep_private_as is false %}
```
Gerrit91, line 148: I think this changes configuration for existing deployments because it introduces a new variable and removes this line when it was not defined?

Gerrit91:
Thanks for re-opening this!

partition/roles/sonic/templates/frr.conf.j2
robertvolkmann: It removes the line if is defined and true. Looks safe for me.

partition/roles/sonic/templates/frr.conf.j2
```diff
@@ -26,12 +26,21 @@ interface {{ port }}
  no ipv6 nd suppress-ra
 {% endfor %}
 {% for k, i in sonic_interconnects.items() %}
+{% if i.vrf is defined %}
+{% for n in i.unnumbered_interfaces|default([]) %}
+!
+interface {{ n }} vrf {{ i.vrf }}
```
robertvolkmann, line 32: Why do we need `vrf` here? I never saw this before.

partition/roles/sonic/templates/frr.conf.j2
mreiger: It renders the `remove-private-AS` statement if the `keep_private_as variable` is not defined or false; existing deployments do not have the variable defined so it should not change anything.

Also it does not break the `exit` test case. :)

partition/roles/sonic/templates/frr.conf.j2
mreiger: With static machine ports we can have interconnects with `unnumbered_interfaces` in a vrf.
I changed the template to make what happens clearer; can you have another look?

mreiger: Interesting! :D

The documentation of frr I found is a bit vague, but it seems to suggest to use the `vrf` statement to add the interface to the vrf it belongs to.
Of course on Sonic there's also the `config_db.json` where this relation is also configured. I would suggest that it is safest to keep the Sonic and frr versions of the truth in sync.

This is also how the `metal-core` configures for regular machine ports; I wanted to replicate this behaviour closely so there is less of a chance for surprises.

robertvolkmann (approved):

partition/roles/sonic/templates/frr.conf.j2
Gerrit91: That's true, I  misread the condition. Thanks for clarification.

Gerrit91 (approved):

PR #418, opened 2025-04-25
Removed unsupported taint keys

robertvolkmann:
## Description

> In previous Gardener versions (< 1.5) these settings were controlled via taint keys (seed.gardener.cloud/{disable-capacity-reservation,invisible}). The taint keys are no longer supported and removed in version 1.12.

See: https://gardener.cloud/docs/gardener/seed_settings/#scheduling

Taint keys were deprecated in https://github.com/gardener/gardener/pull/2315


Gerrit91:
Thanks. Maybe it would be cool to have these seed settings configurable through a variable, too? Was the protected taint also removed as discussed in the referenced pull request?

Gerrit91 (approved):
Will carry this over into the operator PR, too.

PR #421, opened 2025-04-29
Sonic role: Enhancements to support multiple neighbors per interconnect

mreiger:
## Description

Implements changes the Sonic role to enable:
* Multiple .1q subinterfaces on a physical port
* Multiple neighbors in an interconnect with individual routemaps

This is to enable a scenario where an exit router has two upstream peers, one is the preferred and the other is the secondary one. For this purpose, the peers can have different routemaps with individual AS path lengths through the prepend_as statement.

The subinterfaces enable the two peers being on the same physical connection separated with VLAN tags. Using subinterfaces instead of VLAN interfaces with tagged ports make the purpose as external connection versus local VLAN clearer.


Honigeintopf:
Code itself looks good also no breakin changes.
If u tested this in Metro, we can merge it.

partition/roles/sonic/tasks/config_db.yaml
```diff
@@ -52,6 +63,15 @@
     sonic_running_cfg_platform: "{{ sonic_running_cfg | community.general.json_query('DEVICE_METADATA.localhost.platform') }}"
     sonic_running_cfg_ports: "{{ sonic_running_cfg | community.general.json_query('PORT')  }}"
 
+- name: Debug port entries without index
```
Honigeintopf, line 66: Are these debugs needed?

partition/roles/sonic/tasks/config_db.yaml
```diff
@@ -44,6 +44,17 @@
   ansible.builtin.set_fact:
     sonic_running_cfg: "{{ sonic_running_cfg_result.stdout | from_json }}"
 
+# Add this right after extracting the running configuration
```
Honigeintopf, line 47: And this?

iljarotar (changes requested):
With https://github.com/metal-stack/metal-roles/pull/390  a new `sonic-config` role was introduced that is meant to replace the `sonic` role in the future. Could you please add your changes to that role as well?

iljarotar:
Depends on https://github.com/metal-stack/sonic-configdb-utils/pull/20.

iljarotar (approved):

mwennrich (approved):

PR #429, opened 2025-05-28
Replace symlinks by role dependencies.

Gerrit91:
## Description

We are blocked upgrading Ansible due to the symlinks used in this repository, which purpose it is to get some common default variables into the role scope. The issue is that future ansible-galaxy versions do not honor symlinks anymore:

- https://github.com/metal-stack/metal-deployment-base/pull/29
- https://github.com/metal-stack/metal-deployment-base/pull/33

I figured this approach out when preparing the migration to metal-ansible-collections. It works here as well. The idea is to put the defaults into an own role (just holding the defaults, no tasks) and then referencing them as a role dependency through `meta`.

As we are a bit stuck with the metal-ansible-collections project, maybe we should cherry-pick this idea into this project.

The disadvantage of this solution is that the role dependencies work quite implicitly. I think only for the purpose of picking up variables this is fine, if tasks were run this would be a different story. Alternatively, every role that depends on these defaults could explicitly import the variables using the `include_vars` module. I think this would look pretty complex and quirky though. Thus, I prefer this solution for this purpose.


control-plane/roles/auditing-meili/meta/main.yml
```diff
@@ -0,0 +1,17 @@
+---
+galaxy_info:
+  role_name: auditing-meili
```
vknabel, line 3: meili can be removed

vknabel (commented):
Looks reasonable and better than the current symlink approach.

robertvolkmann:
@Gerrit91 could you deploy metal-stack with your changes?

robertvolkmann:
> I hope that [429a3ca](https://github.com/metal-stack/metal-roles/commit/429a3cade6342d54aed7f5ab54907a9bc89733e1) did not destroy something? The reason for nesting the `main.yaml` into `main` is that this recursively picks up all files in this directory, so somethings like `extensions.yaml` would also be picked up. If you just put this into `defaults`, then only `main.yaml` is picked up but no other files next to the `main.yaml`.

I already learned it the hard way.

robertvolkmann (approved):

PR #430, opened 2025-06-05
Add loopback as default src address for BGP

iljarotar:
## Description

Source address is not set correctly if `frr_mgmt_framework_config` and management VRF are both enabled. If management VRF is disabled eth0 address is used as the default source address. But we would rather not disable management VRF. Currently, what happens is described in this [issue](https://github.com/sonic-net/sonic-buildimage/issues/14195). So as a workaround we disable management VRF on all switches that require internet connectivity. This is suboptimal.
If `frr_mgmt_framework_config` is disabled the default source address is the loopback address. This seems reasonable. But instead of relying on this variable I explicitly set the source address in the frr.conf.


robertvolkmann:
Tested it in our starter.

PR #449, opened 2025-07-09
Use `metal-view` as hmac default variable for `metal-metrics-exporter`

simcod:
## Description

The `metal-metrics-exporter` needs a `view` HMAC. This pull request makes this even clearer.

Related to: https://github.com/metal-stack/metal-metrics-exporter/pull/24

### Required Actions

```ACTIONS_REQUIRED
We recommend running the `metal-metrics-exporter` with a `Metal-View` auth type. The new Ansible variable `monitoring_metal_api_authtype` was introduced. By default, `Metal-View` is used. Please make sure, to update the auth type, if you want to run the exporter with higher privileges or verify, if `monitoring_metal_api_hmac` is configured correctly to be a view HMAC.
```


Gerrit91:
This requires a corresponding pull request for the metrics exporter to initialize the client correctly. Also an `ACTIONS_REQUIRED` section would be necessary.

Gerrit91 (approved):

PR #451, opened 2025-07-11
Fix sonic generate environment file not found

iljarotar:
## Description

For some reason `/etc/sonic/sonic-environment` could not be found when `/etc/sonic` was mounted. Copying it to `/tmp/sonic` and mounting that directory instead solved the problem. Maybe I misunderstood the docker module, please advise.


vknabel:
As discussed in a quick call, @iljarotar and I decided that he tries to avoid the copy-to-tmp workaround by fixing the issue. If it takes too much time, this workaround is fine.

PR #453, opened 2025-07-16
Migrate metal-image-cache-sync to V2 API

majst01:
## Description

Part of:

- [ ] https://github.com/metal-stack/metal-image-cache-sync/pull/21

depends on: 

- [x] https://github.com/metal-stack/api/pull/15

### Breaking Change

```BREAKING_CHANGE
The variable `metal_apiserver_session_secret` is now mandatory for security reasons.
```


partition/roles/image-cache/tasks/image-sync.yaml
```diff
@@ -1,4 +1,19 @@
 ---
+- name: Ensure cache directory
+  ansible.builtin.file:
+    path: "{{ image_cache_sync_host_path }}"
+    state: directory
+    mode: '0755'
```
majst01, line 6: Wonder why this is new here, where did this directory got created before?

vknabel: Good question... the metal-image-cache-sync does not create the cache dir

vknabel:
Good to go from my perspective.

vknabel:
We decided on the following changes:

- [x] `metal_deployment_admin_token` no default, but documentation
- [x] `image_cache_sync_deployment_admin_token: "{{ metal_deployment_admin_token }}"`
- [x] `image-cache` role should not check for `metal_deployment_admin_token` anywhere
- [x] no `import_tasks` of `metal-deployment-token` in `image-cache`
- [x] idea: either set `image_cache_sync_deployment_admin_token` or `metal_deployment_admin_token` manually in production
- [x] update documentation accordingly

partition/roles/image-cache/tasks/image-sync.yaml
majst01: on real hardware this is done with the lv-create of a dedicated partition

partition/roles/image-cache/tasks/image-sync.yaml
```diff
@@ -7,13 +22,14 @@
...
+      METAL_IMAGE_CACHE_SYNC_HOSTNAME: "{{ inventory_hostname }}"
       METAL_IMAGE_CACHE_SYNC_ENABLE_KERNEL_CACHE: "{{ image_cache_kernel_cache_enabled }}"
       METAL_IMAGE_CACHE_SYNC_ENABLE_BOOT_IMAGE_CACHE: "{{ image_cache_boot_image_cache_enabled }}"
       METAL_IMAGE_CACHE_SYNC_MAX_CACHE_SIZE: "{{ image_cache_sync_max_cache_size }}"
       METAL_IMAGE_CACHE_SYNC_MAX_IMAGES_PER_NAME: "{{ image_cache_sync_max_images_per_name }}"
       METAL_IMAGE_CACHE_SYNC_MIN_IMAGES_PER_NAME: "{{ image_cache_sync_min_images_per_name }}"
-      METAL_IMAGE_CACHE_SYNC_METAL_API_ENDPOINT: "{{ image_cache_sync_metal_api_endpoint }}"
-      METAL_IMAGE_CACHE_SYNC_METAL_API_HMAC: "{{ image_cache_sync_metal_api_view_hmac }}"
+      METAL_IMAGE_CACHE_SYNC_METAL_APISERVER_URL: "{{ image_cache_sync_metal_apiserver_url }}"
+      METAL_IMAGE_CACHE_SYNC_METAL_APISERVER_TOKEN_PATH: "/var/lib/metal-image-cache-sync/metal-apiserver-token"
```
iljarotar, line 32: shouldn't `/var/lib/metal-image-cache-sync` be configurable if `image_cache_sync_host_path` is?

vknabel: Currently it puts the token into `dest: "{{ image_cache_sync_host_path }}/metal-apiserver-token"`. We either need to use `image_cache_sync_host_path` here (this is fine for me), or allow completely overriding this.

Keep in mind to tell the metal-image-cache-sync the new location if configurable.

Gerrit91: But why does internal container path need to be configurable? I think it's fine when it is just hard-wired as long as on the host system you can decide where the mount needs to be done? See a few lines below that the internal container path is static.

vknabel: Ah, yup, you're right. Hard-coded is fine. Mistook `image_cache_sync_host_path` to exactly be `/var/lib/metal-image-cache-sync`, which is not the case (as you said host vs container). 👍

vknabel (approved):
fine from my perspective 👍

Gerrit91 (approved):

PR #472, opened 2025-09-17
fix: logging seed kubeconfig

vknabel:
## Description

During a deployment we noticed that the way to retrieve the kubeconfig is outdated. This adapts the new way of doing it.


vknabel:
I wanted to apply some refactoring and to now write that file, but `ansible-common/roles/helm-chart` requires the kubeconfig to be a file.

Gerrit91 (commented):
Could work, but I cannot test it easily. :see_no_evil:

Gerrit91 (approved):

PR #476, opened 2025-09-26
Bump monitoring to match current Gardener version.

Gerrit91:
## Description

Supersedes #472.


control-plane/roles/monitoring/defaults/main.yaml
```diff
@@ -1,4 +1,7 @@
 ---
+# TODO: move to release vector
+prometheus_chart_version: 73.2.3
```
Gerrit91, line 3: Has to be addressed before merge.

PR #480, opened 2025-10-08
Fix typos

majst01:
## Description

None


control-plane/roles/gardener-operator/defaults/main/operator.yaml
```diff
@@ -4,7 +4,7 @@ gardener_operator_garden_name: "{{ gardener_defaults_garden_name }}"
 gardener_operator_repo_url: "{{ gardener_repo_url }}"
 gardener_operator_repo_ref: "{{ gardener_repo_ref }}"
 
-gardener_opreator_backup_infrastructure:
+gardener_opereator_backup_infrastructure:
```
majst01, line 7: I dont know if this breaks actual deployments, so be aware

Gerrit91: Probably, but only test systems, so it should be fine. Looks still wrong though. :see_no_evil:

PR #493, opened 2025-11-06
Add roles for deploying Zitadel

ostempel:
## Description

Add role to bootstrap zitadel postgres database.


Gerrit91:
Either the zitadel role goes here or the zitadel-db role goes into the mini-lab, but there is no reason to split them up.

PR #500, opened 2025-11-26
feat(nsq): tls require mode

vknabel:
## Description

None


control-plane/roles/nsq/templates/nsq.yaml
```diff
@@ -51,7 +51,11 @@ spec:
         - -broadcast-address={{ nsq_broadcast_address }}
         - -data-path=/data
 {% if nsq_tls_enabled %}
-        - -tls-required=tcp-https
+{% if nsq_tls_require_mode is boolean %}
+- -tls-required={{ nsq_tls_require_mode | to_json }}
+{% else %}
+- -tls-required={{ nsq_tls_require_mode | default("tcp-https") }}
```
Gerrit91: Default is not required when it's already set in the `defaults/main.yaml`.

PR #504, opened 2025-12-18
Fix readme and some defaults in the sonic-config role

iljarotar:
## Description

depends on https://github.com/metal-stack/sonic-configdb-utils/pull/36

### DIFF
--- a/partition/roles/sonic-config/README.md
+++ b/partition/roles/sonic-config/README.md
@@ -466,6 +466,9 @@ sonic_config_portchannels:
         - Ethernet116
 
 # Ports configuration.
+# If this variable is null, which is default, no port configuration will be generated.
+# Otherwise, the generator will add default configuration for all ports present on the switch.
+# Default values for all ports and for each individual port can be overwritten.
 sonic_config_ports:
   # Default value for auto negotiation.
   default_autoneg: off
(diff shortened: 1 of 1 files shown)


partition/roles/sonic-config/defaults/main.yaml
```diff
@@ -1,5 +1,5 @@
 sonic_config_bgp_ports: []
-sonic_config_breakouts: {}
+sonic_config_breakouts:
```
Gerrit91: Where does this matter? Shouldn't the config generator be able to correctly handle empty dicts as well?

partition/roles/sonic-config/defaults/main.yaml
```diff
@@ -19,7 +19,7 @@ sonic_config_mgmt_vrf: false
 sonic_config_nameservers: []
 sonic_config_ntp: {}
 sonic_config_portchannels: {}
-sonic_config_ports: {}
```
Gerrit91: Same question as with https://github.com/metal-stack/metal-roles/pull/504/changes/BASE..d556b9b178b78034883ff37e12215da37ee69a45. 😅

iljarotar: In this case it's a bit different. `ports` is not a map but a struct with a slice of ports and some other fields. If the slice of ports is empty but the `ports` struct is not nil I'd like the generator to create a default config for all ports that are present on the switch. If `ports` is nil it will not generate any port config at all. 
Of course I could just check if the `ports` struct contains only zero values instead of checking for nil, but what for?

Gerrit91: Thanks!

Gerrit91 (approved):

PR #522, opened 2026-02-02
feat: add envoy gateway to service clusters (for isolated clusters)

mwennrich:
## Description

add envoy gateway to service clusters (for isolated clusters)

- [x] successfully deployed in test and created a working forbidden cluster

### DIFF
--- a/control-plane/roles/isolated-clusters/tasks/registry.yaml
+++ b/control-plane/roles/isolated-clusters/tasks/registry.yaml
@@ -2,7 +2,6 @@
 - name: Create namespace {{ isolated_clusters_registry_namespace }}
   k8s:
     definition:
-      api_version: v1
       kind: Namespace
       metadata:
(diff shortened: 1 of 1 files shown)


control-plane/roles/isolated-clusters/tasks/envoy-gateway.yaml
```diff
@@ -0,0 +1,45 @@
...
+    kubeconfig: "{{ cluster.kubeconfig }}"
+    wait: yes
+    wait_condition:
+      type: Available
+      status: "True"
+    wait_timeout: 300
+
+- name: Deploy gateway resources
+  k8s:
+    definition: "{{ lookup('template', 'envoy-gateway/gateway.yaml') }}"
```
Gerrit91: Theoretically, the namespace can be passed here, too and then omitted from all the resources in the file, but that's not so important.

PR #536, opened 2026-02-23
Set time if unsynchronised so that initial pull of sonic-configdb-utils will not fail

### LINKED ISSUE #535: Initial deployment of switches using sonic-config role may fail
### Description

Initial deployment of sonic switches may fail if the internal clock of the switch is so far off that the `sonic-configdb-utils` container can not be pulled.

Workaround: Set the time manually before starting the deployment job.

Suggested solution: Modify the `sonic-config` role to check before pulling the `sonic-configdb-utils` container whether ntp is in sync, and set the time if it is not.

mreiger:
## Description

None

<!--
If possible, please reference other issues or pull requests.

Closes #535.


iljarotar (approved):
works in test

PR #563, opened 2026-03-19
Make default src address loopback optional on SONiC

iljarotar:
## Description

With https://github.com/metal-stack/metal-core/pull/187 metal-core allows configuring whether it should set the default source address to loopback or not.

TODO:

- [x] also allow this configuration for statically rendered `frr.conf` in `sonic-config` role


Gerrit91:
If this is breaking please add a REQUIRED_ACTIONS section or BREAKING_CHANGE section to the release notes.

Gerrit91 (approved):

PR #566, opened 2026-03-23
add xdr provider config if extension is enabled

ulrichSchreiner:
Add the XDR extension if it is configured in the config spec.


control-plane/roles/gardener-shoots/templates/shoot.yaml.j2
```diff
@@ -68,8 +68,16 @@ spec:
       nodeCIDRMaskSize: {{ _config.node_cidr_mask_size }}
     verticalPodAutoscaler:
       enabled: {{ _config.vertical_pod_autoscaler }}
-{% if 'audit_extension_splunk' in _config or 'csi_driver_lvm_extension' in _config %}
+{% if 'audit_extension_splunk' in _config or 'csi_driver_lvm_extension' in _config or 'xdr_extension' in _config %}
```
Gerrit91, line 71: Can you please add an example configuration to the defaults, like?

https://github.com/metal-stack/metal-roles/blob/master/control-plane/roles/gardener-shoots/defaults/main.yaml#L74-L76

Gerrit91 (approved):

PR #570, opened 2026-03-30
Sonic config prefix lists

iljarotar:
## Description

This allows adding additional prefix lists. Currently prefix lists can only be defined for additional peer groups but not for the default `FABRIC` peer group.


muhittink (approved):
LGTM

PR #585, opened 2026-04-20
Add param for deploying `ClusterRoleBinding`s into the virtual garden.

Gerrit91:
## Description

This allows providing some cluster role bindings during the deployment into the virtual garden, e.g. for creating access for OIDC users.

/cc @simcod 

#### Used AI-Tools ✨

- None used for generation


simcod (approved):

Gerrit91:
As this implementation was so generic, I think we could also just call the variable `gardener_virtual_garden_access_extra_resources` or something instead of `gardener_virtual_garden_access_cluster_role_bindings`. But now it's too late. :sweat_smile:

PR #592, opened 2026-04-29
chore: replace partition promtail with alloy

ma-hartma:
## Description

Replaces promtail with alloy in the partition.
There is an own PR for the control-plane replacement: https://github.com/metal-stack/metal-roles/pull/595.

- [x] deploy alloy
    - [x] role for docker service in partition (switches and management servers)
- [x] convert existing configs
   - [x] convert existing scrape configs with alloy convert and re-templated them
   - [x] provide as snippets
   - [x] multiple loki targets
   - [x] timeouts and basic auth for loki targets
- [x] logs and metrics for alloy itself (https://grafana.com/docs/alloy/latest/collect/metamonitoring/=) to loki and prometheus
   - [x] partition
   - [x] document prometheus_alloy targets
 - [x] positions file
 - [x] old labels 
 
Question Marks: 
 - [ ] prometheus inside the partition could be replaced with alloy completely as it can pick up the metrics as well, but aggregation would have to be re-evaluated 

After successful replacement
- [x] update monitoring docs for the website
- [x] migration guide 
    - [x] (mention https://grafana.com/docs/alloy/latest/set-up/migrate/from-promtail/#limitations)
- [x] mini-lab uses alloy instead of promtail (WIP https://github.com/metal-stack/mini-lab/pull/298)
- [x] deprecate promtail roles

Manually tested in mini-lab:
- [x] syslog snippet
- [x] syslog snippet with positions file
- [x] journal snippet
- [x] docker snippet
- [x] journal-file snippet with /run/log/journal and /var/log/journal 
- [x] journal-file snippet with positions file, missing positions name and existing positions name
- [x] alloy-meta snippet
- [x] metrics are collected via prometheus
- [x] labels as expected (logs and alloy metrics)
- [x] cutover migration from promtail to alloy in the mini-lab with existing positions files and stopping promtail 

References:

- https://github.com/metal-stack/metal-roles/pull/595 control-plane promtail->alloy migration
- https://github.com/metal-stack/mini-lab/pull/298 to test this in the mini-lab
- https://github.com/metal-stack/website/pull/260 website docs update
- https://github.com/metal-stack/releases/pull/287 add image to releases
- https://github.com/metal-stack/metal-roles/pull/584 will have to replace promtail as well

Closes https://github.com/metal-stack/metal-roles/issues/552

#### Used AI-Tools ✨

- Claude Sonnet 4.6 for docs and migration guide generation and jinja templating

### Required Actions

```ACTIONS_REQUIRED
[...]

PR #584, opened 2026-04-20
Sflow integration

Sven-Ric:
## Description

Closes #549.

This branch implements sflow based flow monitoring. 

#### Used AI-Tools ✨

claude was used for weathermap visualisation

- claude used for generation

PR #595, opened 2026-05-04
chore: replace control-plane promtail with alloy

ma-hartma:
## Description

Replaces promtail with alloy in the control-plane.
The partition replacement is treated separately in https://github.com/metal-stack/metal-roles/pull/592 and should go first.

- [x] deploy alloy
    - [x] role for helm chart in k8s cluster
- [x] convert existing configs
   - [x] pods
   - [x] events
   - [x] labels
   - [x] metrics
- [x] logs and metrics for alloy itself (https://grafana.com/docs/alloy/latest/collect/metamonitoring/=) to loki and prometheus
 - [x] WAL persistence to allow restarts
 - [x] check promtail helm charts, as they had lots of defaults that alloy doesn't: https://github.com/grafana/helm-charts/tree/main/charts/promtail
  
Question Marks: 
 - [x] evaluate using https://github.com/grafana/k8s-monitoring-helm instead of https://github.com/prometheus-operator/kube-prometheus

After successful replacement
- [x] update monitoring docs for the website
- [x] migration guide 
    - [x] convert does not work with helm chart
- [x] deprecate promtail roles

Manually tested in the mini-lab and demo-rack
- [x] metrics are pushed to thanos, accessible via Grafana
- [x] labels work as documented and before
- [x] cutover migration from promtail to alloy in the mini-lab including cleanup
- [x] deprecation warnings show up

Still needs to be tested:
- [x] enable Service Monitors

References:

- https://github.com/metal-stack/metal-roles/pull/592
- https://github.com/metal-stack/website/pull/260 website docs update
- https://github.com/metal-stack/releases/pull/288
- https://github.com/metal-stack/metal-roles/pull/584 will have to replace promtail as well

Closes https://github.com/metal-stack/metal-roles/issues/552

#### Used AI-Tools ✨

- Claude Sonnet 4.6 for docs and migration guide generation and jinja templating

### Breaking Change

```BREAKING_CHANGE
The event `job` label was renamed from `monitoring/event-exporter` to `events`. Update any existing LogQL queries, dashboard filters, and alert rules that reference `{job="monitoring/event-exporter"}` to `{job="events"}`.

The monitoring role's basic auth configuration for `Thanos Receive Ingress` has changed. The raw `htpasswd` string `monitoring_thanos_receive_ingress_basic_auth` has been replaced by two plaintext variables `
monitoring_thanos_receive_ingress_basic_auth_user: thanos-receive` and `monitoring_thanos_receive_ingress_basic_auth_password: mysecret`.
See: https://github.com/metal-stack/metal-roles/blob/master/control-plane/roles/monitoring/README.md#thanos-receive-ingress-credentials
```

### Required Actions

```ACTIONS_REQUIRED
The control-plane `logging` role now deploys Grafana `Alloy` as the default log collector. Alloy needs to be configured before running the new version of the role, as the deprecated `Promtail` DaemonSet and the `event-exporter` resources are disabled and removed by default.

Follow the migration instructions:
- https://github.com/metal-stack/metal-roles/blob/master/control-plane/roles/logging-common/README.md#migration-from-promtail

Several migration scenarios are supported — from a hard cutover to a parallel run for verification before switching over. Note that running both simultaneously will produce duplicate log entries in Loki during the transition window. 
```


control-plane/roles/logging-common/templates/alloy-config.alloy.j2
```diff
@@ -0,0 +1,163 @@
...
+  }
+}
+
+loki.source.kubernetes "pods" {
+  targets    = discovery.relabel.pods.output
+  forward_to = [loki.write.default.receiver]
+}
+
+loki.source.kubernetes_events "events" {
+  job_name   = "kubernetes-events"
```
ma-hartma, line 116: The event-exporter used with promtail had the events labelled with `job=monitoring/event-exporter`.
Should we actually re-label the logs from `job=kubernetes-events` to also match the legacy job name?
I haven't found any usage in metal-roles, but other users may rely on it.

control-plane/roles/logging/defaults/main.yaml
```diff
@@ -1,5 +1,32 @@
...
+# max_keepalive_time: samples older than this are dropped even if not yet delivered (default: Alloy built-in 8h)
+logging_alloy_prometheus_wal_truncate_frequency: "2h"
+logging_alloy_prometheus_wal_max_keepalive_time: "8h"
+
+# Full Alloy River config override. When set, bypasses all structured vars above.
+# logging_alloy_config_raw: |
+
 logging_ingress_dns: "loki.{{ metal_control_plane_ingress_dns }}"
 logging_ingress_loki_tls: yes
 logging_ingress_loki_basic_auth_user: promtail
```
ma-hartma, line 50: I'll leave the old username for now.

control-plane/roles/logging-common/templates/alloy-config.alloy.j2
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  rule {
+    target_label = "cluster"
+    replacement  = "{{ logging_alloy_cluster_label }}"
+  }
+
+  rule {
+    target_label = "job"
+    replacement  = "monitoring/event-exporter"
```
ma-hartma: The event-exporter used with promtail had the events labelled with job=monitoring/event-exporter.
I did actually re-label the logs from job=kubernetes-events to also match the legacy job name.
I haven't found any usage in metal-roles, but other users may rely on it.

control-plane/roles/logging/defaults/main.yaml
```diff
@@ -1,8 +1,45 @@
...
+# Set to true to also deploy Promtail alongside Alloy during a migration window.
+# Both DaemonSets will ship logs simultaneously — Loki will receive duplicate entries.
+# Requires logging_promtail_chart_version and logging_promtail_chart_repo to be set.
+# See the Migration from Promtail section in README.md.
+# logging_promtail_enabled: false
+
 logging_ingress_dns: "loki.{{ metal_control_plane_ingress_dns }}"
 logging_ingress_loki_tls: yes
-logging_ingress_loki_basic_auth_user: promtail
+logging_ingress_loki_basic_auth_user: promtail # TODO rename to alloy or generic
```
ma-hartma: I'll leave the old username for now.

control-plane/roles/logging/defaults/main.yaml
```diff
@@ -1,8 +1,45 @@
...
+
+# Value for the cluster= label attached to all log and metric streams via relabel rules
+logging_alloy_cluster_label: "{{ metal_control_plane_stage_name }}"
+
+# Prometheus remote_write endpoints for Alloy self-metrics (prometheus.exporter.self).
+# Alloy scrapes its own /metrics and pushes them via remote_write.
+# Each entry: {url, remote_timeout?: duration, basic_auth?: {username, password}}
+# Thanos Receive endpoint is included automatically when monitoring_thanos_receive_enabled: true.
+logging_alloy_prometheus_write_endpoints: >-
+  {{ [{'url': 'http://thanos-receive.' ~ logging_namespace ~ '.svc.cluster.local:19291/api/v1/receive'}]
```
ma-hartma: In the shooted seeds, we do not have an in-cluster prometheus. Here we could actually use ServiceMonitors but I decided against it, because we would introduce a dependency on prometheus already existing before deploying the helm chart.

control-plane/roles/logging/defaults/main.yaml
```diff
@@ -1,7 +1,46 @@
...
 logging_namespace: monitoring
+logging_alloy_port: 12345
+
+# Alloy deployment (default false). Set true to deploy Alloy.
+# Requires logging_alloy_chart_version and logging_alloy_chart_repo when enabled.
+logging_alloy_enabled: false
+
+# Loki push endpoints.
+logging_alloy_loki_write_endpoints:
+  - url: "http://loki:3100/loki/api/v1/push"
```
ma-hartma: We decided not to go with a default here, as it is hard to unify for all envs.

control-plane/roles/logging/defaults/main.yaml
```diff
@@ -1,7 +1,46 @@
...
+
+# Value for the cluster= label attached to all log and metric streams via relabel rules
+logging_alloy_cluster_label: "{{ metal_control_plane_stage_name }}"
+
+# Prometheus remote_write endpoints for Alloy self-metrics (prometheus.exporter.self).
+# Alloy scrapes its own /metrics and pushes them via remote_write.
+# Each entry: {url, remote_timeout?: duration, basic_auth?: {username, password}}
+# Thanos Receive endpoint is included automatically when monitoring_thanos_receive_enabled: true.
+logging_alloy_prometheus_write_endpoints: >-
+  {{ [{'url': 'http://thanos-receive.' ~ logging_namespace ~ '.svc.cluster.local:19291/api/v1/receive'}]
```
ma-hartma: We decided not to go with a default here, as it is hard to unify for all envs.

simcod:
Successfully tested the migration from promtail to alloy for the `monitoring` and `logging` roles.

ma-hartma:
The `gardener-logging` role update is now in a seperate branch, as it needs to be re-evaluated and tested separately: https://github.com/metal-stack/metal-roles/pull/622.

control-plane/roles/logging/defaults/main.yaml
```diff
@@ -1,7 +1,54 @@
...
+logging_alloy_prometheus_wal_max_keepalive_time: "8h"
+
+# Full Alloy River config override. When set, bypasses all structured vars above.
+# logging_alloy_config_raw: |
+
+# Promtail deployment (default true, deprecated). Enabled by default for backward
+# compatibility with existing installs. Set false once migrated to Alloy.
+# Requires logging_promtail_chart_version and logging_promtail_chart_repo when enabled.
+# See the Migration from Promtail section in README.md.
+logging_promtail_enabled: true
```
Gerrit91: Can we default this the other way around? Such that alloy gets deployed and promtail gets uninstalled by default when executing this role? When the alloy deployment succeeds then just assume everything is alright. Ideally, it's just the operator running the role in the pipeline once and everything is done. These logs are not audit logs, if in case of something going wrong we lose something, it is not catastrophic and operators can fix it right away.

control-plane/roles/logging/tasks/main.yaml
```diff
@@ -43,3 +71,11 @@
...
     helm_chart_version: "{{ logging_promtail_chart_version }}"
     helm_value_file_template: "promtail-values.yaml"
+  when: logging_promtail_enabled
+
+- name: Remove Promtail
+  kubernetes.core.helm:
+    name: promtail
+    namespace: "{{ logging_namespace }}"
+    state: absent
+  when: logging_promtail_migrate_cleanup | default(false) | bool
```
Gerrit91: Set in defaults and use `not logging_promtail_enabled` as default?

control-plane/roles/logging-common/templates/alloy-config.alloy.j2
Gerrit91: You can also rename if you want.

control-plane/roles/logging/templates/alloy-values.yaml
```diff
@@ -0,0 +1,47 @@
...
+alloy:
+  configMap:
+    # -- Create a new ConfigMap for the config file.
+    create: true
+    # -- Content to assign to the new ConfigMap.  This is passed into `tpl` allowing for templating from values.
+    content: |-
+      {{ logging_alloy_config | indent(6) }}
+
+  # -- Port to listen for traffic on.
+  listenPort: {{ logging_alloy_port }}
```
Gerrit91: Can't we just use the default?

control-plane/roles/logging/README.md
```diff
@@ -1,31 +1,144 @@
...
-The role includes tasks to install and configure the following logging tools:
+Deploys the control-plane logging stack into the Kubernetes control-plane cluster.
 
-- Loki
-- Logging ingress for Loki
-- Promtail for monitoring the control plane cluster
+Components:
+
+- **Loki** — log storage and query backend
+- **Alloy** — log collector (DaemonSet), collects pod logs via the Kubernetes API (`loki.source.kubernetes`) and forwards them to Loki
```
Gerrit91: I think if we use the Kubernetes collector the DaemonSet would be overkill for large clusters? Is it really good to use it and not just give the pods mount permissions as before? I can imagine this is quite heavy on the network for productive workloads?

control-plane/roles/logging/README.md
```diff
@@ -1,31 +1,144 @@
...
+    basic_auth:
+      username: "{{ monitoring_thanos_receive_ingress_basic_auth_user }}"
+      password: "{{ monitoring_thanos_receive_ingress_basic_auth_password }}"
+```
+
+See the [monitoring role migration guide](../monitoring/README.md#thanos-receive-ingress-credentials) for credential setup. Do not set both options simultaneously — `logging_alloy_service_monitor_enabled` and `logging_alloy_prometheus_write_endpoints` are mutually exclusive.
+
+### Logs
+
+Alloy runs as a Kubernetes DaemonSet, so its own pod logs are captured by `loki.source.kubernetes` automatically — no additional configuration is needed.
```
Gerrit91: But this is true for StatefulSets or Deployments, too.

control-plane/roles/monitoring/defaults/main.yaml
```diff
@@ -70,7 +73,9 @@ monitoring_thanos_receive_enabled: false
 monitoring_thanos_receive_ingress_enabled: false
 monitoring_thanos_receive_ingress_dns: "thanos-receive.{{ metal_control_plane_ingress_dns }}"
 monitoring_thanos_receive_ingress_annotations: []
-monitoring_thanos_receive_ingress_basic_auth: {}
+monitoring_thanos_receive_ingress_basic_auth_user: thanos-receive
+monitoring_thanos_receive_ingress_basic_auth_password:
+monitoring_thanos_receive_ingress_basic_auth_password_salt: "{{ monitoring_thanos_receive_ingress_basic_auth_password | hash('sha1') | truncate(22, True, '') }}"
```
Gerrit91, line 78: Is this breaking change required or can we postpone and put it into another PR?

control-plane/roles/logging/defaults/main.yaml
ma-hartma: I like that and go for it! I was just careful and had the feeling an opt-in would be favorable, because the partition migration does.

control-plane/roles/monitoring/defaults/main.yaml
ma-hartma: I only changed this, because we want to use push-based metrics for Alloy in gardener-logging and I need the credentials separated there. Also: all other ingresses in the logging/monitoring roles use the format that I have just updated to. What's your recommendation?

control-plane/roles/logging/README.md
ma-hartma: Very interesting thought. I hadn't considered this and went with Alloy's default (DaemonSet), because we also have Promtail deployed as a DaemonSet. I only know now: for Promtail there was actually a Reason: It uses the Kubernetes API **and additionally** reads the Pod logs from the filesystem on every host. 

My new implementation chose the simplest way, just querying the Kubernetes API for pod logs and events. 

The requirements for low-workload and high-workload clusters would actually differ I think:

**Small-Medium Size Cluster**
We could go with the Deployment and only query the API, the cluster can handle the extra load on it.

**Clusters with lots of containers**
A DaemonSet querying /var/log/pods/... instead of the Kubernetes API for pod logs. Only get the events from the API. 
Would not put all load on the API, but also on all the single hosts.

Do you think, gardener-logging and control-plane logging should go with different setups or can we decide on one?

control-plane/roles/logging-common/templates/alloy-config.alloy.j2
ma-hartma: Do you prefer "events" or "kubernetes-events"?

control-plane/roles/logging/README.md
ma-hartma: Also to consider:
If we just use the **Deployment** with querying the Kubernetes API - and Alloy goes down, we lose pod logs and events for that time window. The alerts could help that we at least find out about that: https://github.com/metal-stack/metal-roles/pull/623
--> Here we wouldn't even need a volume at all, because persisting positions files for API requests makes no sense.

If we use the **DaemonSet** instead, we are not in danger of losing pod logs, even if the API is unreachable or Alloy down for some time.
But we then need a Volume to persist our positions on the watched files.

Gerrit91: What is the issue with just mounting the directories on the hosts and using DaemonSet? Is it just extended permissions? Doesn't this outweigh that positions can be stored? From the complexity it does not sound so complicated to me to add this mount, I think it's even prepared in the chart?

ma-hartma: There is no issue with the implementation at all. I was just unsure which one was the better fit for our use-case. 

My proposal would then be, to go with DaemonSets and the file reads on /var/log/pods.
- Yes, it is extended permissions, but the same as promtail always had
- It takes load off the API
- It is easier to track position files on local files, making this more resilient for the pod logs at least and support container restarts without re-delivery or losing logs

To avoid shipping Kubernetes events (which still have to be queried from the API)  multiple times, because all the pods of the DaemonSet want to query it, I also need to add clustering then.

control-plane/roles/monitoring/defaults/main.yaml
Gerrit91: We can keep it. But the release notes will be quite long and we must make sure this will not be overseen, maybe put it first in the release notes.

control-plane/roles/monitoring/tasks/main.yaml
```diff
@@ -9,11 +9,20 @@
...
+- name: Check deprecated variables are not set
+  assert:
+    fail_msg: >
+      'monitoring_thanos_receive_ingress_basic_auth' has been removed.
+      Please migrate to 'monitoring_thanos_receive_ingress_basic_auth_user' and
+      'monitoring_thanos_receive_ingress_basic_auth_password' — the htpasswd entry
+      is now generated automatically. See the Migration section in the role README.
+    quiet: true
+    that:
+      - monitoring_thanos_receive_ingress_basic_auth is not defined or monitoring_thanos_receive_ingress_basic_auth == {}
```
ma-hartma, line 24: @Gerrit91 I added a guard, so this should be fine with the additional release notes?

control-plane/roles/monitoring/defaults/main.yaml
ma-hartma: I think we are quite safe with this guard: https://github.com/metal-stack/metal-roles/pull/595#discussion_r3403793728

control-plane/roles/logging-common/defaults/main.yaml
```diff
@@ -0,0 +1,41 @@
+---
+# Alloy target namespace.
+logging_alloy_namespace: monitoring
```
Gerrit91: Prefix with `logging_common_`

control-plane/roles/logging-common/tasks/deploy-alloy.yaml
```diff
@@ -0,0 +1,25 @@
+---
+- name: Check mandatory variables for logging-common are set
```
Gerrit91, line 5: Add task running `setup_yaml` module and add `ansible-common`  to role dependencies.

Gerrit91 (approved):

PR #599, opened 2026-05-22
Deploy gardener metrics exporter again

simcod:
## Description

Deploys the `gardener-metrics-exporter` again within our `monitoring` role.

#### Used AI-Tools ✨

- None used for generation


control-plane/roles/monitoring/tasks/exporters.yaml
```diff
@@ -17,10 +17,27 @@
...
+      apiVersion: v1
+      kind: Secret
+      metadata:
+        name: seed-kubecfg
+        namespace: "{{ monitoring_namespace }}"
+      stringData:
+        kubecfg.yaml: "{{ virtual_garden_kubeconfig }}"
+  when: monitoring_gardener_enabled
+
+- name: Deploy gardener-metrics-exporter
```
Gerrit91, line 72: We need to add the kubeconfig-refresher because in many cases this does not run in the same cluster as the Gardener. 🙈 

Here is an inspiration how to do it: https://github.com/metal-stack-cloud/ansible-roles/pull/69

control-plane/roles/monitoring/defaults/main.yaml
```diff
@@ -50,9 +50,18 @@ monitoring_rethinkdb_exporter_metal_db_endpoint: metal-db.metal-control-plane:28
...
+virtual_garden_kubeconfig_refresher_gcp_project_id:
+virtual_garden_kubeconfig_refresher_gcp_location:
+virtual_garden_kubeconfig_refresher_gcp_cluster_name:
+virtual_garden_kubeconfig_refresher_metal_stack_cloud_enabled: false
+virtual_garden_kubeconfig_refresher_metal_stack_cloud_project_id:
+virtual_garden_kubeconfig_refresher_metal_stack_cloud_api_token:
+virtual_garden_kubeconfig_refresher_metal_stack_cloud_cluster_id:
+virtual_garden_kubeconfig_refresher_garden_cluster_enabled: false
+virtual_garden_kubeconfig_refresher_garden_cluster_kubeconfig:
+
```
Gerrit91: Must be prefixed with the role name.

control-plane/roles/monitoring/templates/virtual-garden-kubeconfig-refresher.yaml
```diff
@@ -0,0 +1,156 @@
...
+          value: "/etc/virtual-garden-kubeconfig/kubeconfig"
+{{- if virtual_garden_kubeconfig_refresher_gcp_enabled }}
+        - name: GOOGLE_APPLICATION_CREDENTIALS
+          value: /gcp/gcp_service_account.json
+        - name: GOOGLE_PROJECT_ID
+          value: "{{ virtual_garden_kubeconfig_refresher_gcp_project_id  }}"
+        - name: GOOGLE_LOCATION
+          value: "{{ virtual_garden_kubeconfig_refresher_gcp_location }}"
+        - name: GOOGLE_CLUSTER_NAME
+          value: "{{ virtual_garden_kubeconfig_refresher_gcp_cluster_name }}"
```
Gerrit91: These must become optional configuration options depending on the environment this is getting deployed to.

control-plane/roles/monitoring/templates/virtual-garden-kubeconfig-refresher.yaml
```diff
@@ -0,0 +1,156 @@
...
+  kind: Role
+  name: virtual-garden-kubeconfig-refresher
+  namespace: {{ monitoring_namespace }}
+subjects:
+- kind: ServiceAccount
+  name: virtual-garden-kubeconfig-refresher
+  namespace: {{ monitoring_namespace }}
+---
+# if running in same cluster as garden cluster to read
+apiVersion: rbac.authorization.k8s.io/v1
```
Gerrit91: Probably also add a flag in order not to deploy these RBACs in every scenario.

control-plane/roles/monitoring/templates/virtual-garden-kubeconfig-refresher.yaml
simcod: You are right. They should only be necessary, if `virtual_garden_kubeconfig_refresher_gcp_enabled` is enabled.

control-plane/roles/monitoring/defaults/main.yaml
simcod: Oh, right. Forgot that.

control-plane/roles/monitoring/templates/virtual-garden-kubeconfig-refresher.yaml
simcod: I added the flag: `monitoring_virtual_garden_kubeconfig_refresher_garden_cluster_local_rbac`

control-plane/roles/monitoring/templates/virtual-garden-kubeconfig-refresher.yaml
```diff
@@ -0,0 +1,157 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: virtual-garden-kubeconfig-refresher
+  namespace: {{ monitoring_namespace }}
+subjects:
+- kind: ServiceAccount
+  name: virtual-garden-kubeconfig-refresher
+  namespace: {{ monitoring_namespace }}
+---
+{% if monitoring_virtual_garden_kubeconfig_refresher_garden_cluster_local_rbac %}
```
Gerrit91: Das ist unterm Strich falsch.


```suggestion
{% if monitoring_virtual_garden_kubeconfig_refresher_garden_cluster_local_rbac %}
---
```

control-plane/roles/monitoring/defaults/main.yaml
```diff
@@ -52,9 +52,19 @@ monitoring_rethinkdb_exporter_metal_db_endpoint: metal-db.metal-control-plane:28
 
 # gardener
 monitoring_gardener_enabled: "{{ True if monitoring_gardener_seeds | length > 0 else False }}"
-monitoring_gardener_virtual_garden_kubeconfig: "{{ lookup('k8s', api_version='v1', kind='Secret', namespace='garden', resource_name='garden-kubeconfig-for-admin').get('data',
-  {}).get('kubeconfig') if monitoring_gardener_enabled else None }}"
 monitoring_gardener_seeds: []
+monitoring_virtual_garden_kubeconfig_refresher_gcp_enabled: false
```
Gerrit91: Nit: It's only possible to select one of these "backends" exclusively (depending on order of a switch condition in the code, which one wins, I guess). It would be a bit cleaner to have a variable like:

```yaml
monitoring_virtual_garden_kubeconfig_refresher_cluster_provider: gcp|metalstackcloud|static-kubeconfig`
```

simcod: Thank you for the review!
You are right. I can add this to the PR. Do you think `stack-kubeconfig` is better to understand, instead of `garden-cluster`? If so, I would change the naming in the pull request and also in the `virtual-garden-kubeconfig-refresher` component.

Gerrit91: I am not sure if `garden-cluster` is a good name because the terms should describe the platform where the garden cluster is hosted or how to retrieve a kubeconfig for accessing the Garden cluster. Maybe just kubeconfig would also be fine.

simcod: You are right. I changed it to `static`. In my opinion this makes it easiest to understand. I also changed the downstream environment variable in the `virtual-garden-kubeconfig-refresher` accordingly, see: https://github.com/metal-stack/virtual-garden-kubeconfig-refresher/pull/7

control-plane/roles/monitoring/defaults/main.yaml
```diff
@@ -52,9 +52,20 @@ monitoring_rethinkdb_exporter_metal_db_endpoint: metal-db.metal-control-plane:28
 
 # gardener
 monitoring_gardener_enabled: "{{ True if monitoring_gardener_seeds | length > 0 else False }}"
-monitoring_gardener_virtual_garden_kubeconfig: "{{ lookup('k8s', api_version='v1', kind='Secret', namespace='garden', resource_name='garden-kubeconfig-for-admin').get('data',
-  {}).get('kubeconfig') if monitoring_gardener_enabled else None }}"
 monitoring_gardener_seeds: []
+monitoring_virtual_garden_kubeconfig_refresher_cluster_provider: gcp
```
ma-hartma: As we have different options, I would set no default but leave this to the user.
This would go better with your Check `virtual-garden-kubeconfig-refresher cluster_provider is valid`, as users would notice earlier if they forgot to configure something and provide their extra variables like cluster_name or cluster_id accordingly.

control-plane/roles/monitoring/tasks/main.yaml
```diff
@@ -12,6 +12,16 @@
...
       - gardener_metrics_exporter_image_tag is defined
 
+- name: Check virtual-garden-kubeconfig-refresher cluster_provider is valid
+  ansible.builtin.assert:
+    fail_msg: >
+      'monitoring_virtual_garden_kubeconfig_refresher_cluster_provider' must be
+      one of: gcp, metalstackcloud, static. Actual value: '{{ monitoring_virtual_garden_kubeconfig_refresher_cluster_provider }}'
+    quiet: true
+    that:
+      - monitoring_virtual_garden_kubeconfig_refresher_cluster_provider in ['gcp', 'metalstackcloud', 'static']
```
ma-hartma, line 23: This will then also fail, if there is no default - forcing the user to look into it.

control-plane/roles/monitoring/defaults/main.yaml
Gerrit91: Done.

Gerrit91 (approved):

PR #603, opened 2026-05-26
Stabilize `frr-reload` on sonic

iljarotar:
## Description

This fixes two FRR-related problems:

- [x] Retry `frr-reload` once. Restarting `frr-reload` often fails on first attempt. Most of the times, the second attempt succeeds.
- [x] Only restart `frr-reload` if an `frr.conf` file exists


iljarotar:
Works in test environment and fixes the targeted issues.

muhittink (approved):
LGTM

PR #606, opened 2026-05-28
fix: gardener managed thanos receive ingress tls

ma-hartma:
## Description

Gardener was not able to generate certs for the Thanos Receive ingress, because the secret name and host were missing.
We had only used tls with cert-manager via annotations but not gardener annotations before.
This PR introduces an optional flag to set secret name and host via the Thanos helm chart, for Gardener to create the certificates for the Thanos Receive ingress.

Please merge after https://github.com/metal-stack/metal-roles/pull/592 and https://github.com/metal-stack/metal-roles/pull/595/


ma-hartma:
Was tested successfully with @simcod on the demo-rack.

Gerrit91 (approved):

PR #609, opened 2026-06-01
Demo rack staging

Sven-Ric:
This branch serves as a staging in case multiple branches need to be tested in parallel.


majst01:
Why so much code ? Seems like this is your sflow branch plus demo rack staging code ?

Sven-Ric:
Yes these are the sflow changes and preexisting demo-rack-specific changes. @ma-hartma would also like to deploy features from his metal-roles branch. This is a workaround for the global metal-roles override.

mwindower:
@Sven-Ric please just keep the code for demo rack staging here and remove the rest.

Sven-Ric:
We can delete the branch. This was needed once to test multiple branches at the same time.

PR #622, opened 2026-06-08
chore: replace gardener promtail with alloy

ma-hartma:
## Description

This was split from #595, as it needs to be re-evaluated and tested separately. 

### TODOs
- [x] event-exporter unneeded
- [x] sensible defaults
- [x] test it

#### Used AI-Tools ✨

- Claude Sonnet 4.6 for docs updates

Closes https://github.com/metal-stack/metal-roles/issues/552

### Required Actions

```ACTIONS_REQUIRED
The control-plane `gardener-logging` role now supports Grafana Alloy as the log collector. Alloy needs to be configured before running the new version of the role, as the deprecated `Promtail` DaemonSet is disabled and removed by default.

Follow the migration instructions:
- https://github.com/metal-stack/metal-roles/blob/master/control-plane/roles/logging-common/README.md#migration-from-promtail

Several migration scenarios are supported — from a hard cutover to a parallel run for verification before switching over. Note that running both simultaneously will produce duplicate log entries in Loki during the transition window. 
```


ma-hartma:
@Gerrit91 This would be ready for a test and is also documented in https://github.com/metal-stack/website/pull/260 now.

Gerrit91 (approved):

PR #623, opened 2026-06-10
feat: add alerts for alloy

ma-hartma:
## Description

Just added some alerts for alloy, to avoid logs being lost without us noticing.
Took a look at the alloy [metrics](https://github.com/grafana/alloy/blob/main/internal/component/common/loki/client/metrics.go) and tried to be not too noisy, but catch the relevant errors. 

#### Used AI-Tools ✨

- Claude Sonnet 4.6


ma-hartma:
image

I have tested the alerts in the mini-lab by making the **Loki** unavailable. 
I think that it is quite critical if logs get dropped permanently, that's why they fire quite quickly. 
If it is too noisy, we can adjust it.

control-plane/roles/monitoring/templates/prometheus-stack-values.yaml
```diff
@@ -478,6 +478,46 @@ additionalPrometheusRulesMap:
...
+          severity: critical
+        annotations:
+          summary: "Alloy is dropping log entries due to rate limiting on {{ $labels.instance }}"
+          description: "Alloy on {{ $labels.instance }}{{ if $labels.partition }} (partition: {{ $labels.partition }}){{ else }} (cluster: {{ $labels.cluster }}){{ end }} is dropping entries due to {{ $labels.reason }}. Raise Loki ingestion limits or scale the Loki write tier."
+      # Push failing is diagnostic context for why drops are occurring; single alert suffices.
+      - alert: AlloyLokiPushFailing
+        expr: rate(loki_write_request_duration_seconds_count{status_code!~"2.."}[5m]) > 0
+        for: 5m
+        labels:
+          severity: critical
```
Sven-Ric: If AlloyLokiPushFailing is primarily meant as diagnostic support for the other alerts, wouldn't a warning be fine here?

PR #626, opened 2026-06-15
fix(mgmt-server): use mgmt_server_router_id in frr.conf template

GeertJohan:
It seems there is missing an underscore here but I'm not sure how it's gone unnoticed for [3 years](https://github.com/metal-stack/metal-roles/blame/92134129535624d1dc69c7b9ffe9039312e65110/partition/roles/mgmt-server/templates/frr.conf.j2#L17)?
Or perhaps `routerid` without underscore is preferred, but then [three other files](https://github.com/search?q=repo%3Ametal-stack%2Fmetal-roles+router_id&type=code) err by having the underscore.

### Breaking Change

```BREAKING_CHANGE
In the `mgmt-server` role the misspelled `mgmt_server_routerid` is renamed to `mgmt_server_router_id`. Update your variables accordingly. The role will error on missing variable.
```


iljarotar:
In our deployments we use `mgmt_server_routerid` so this change would be breaking. I'd prefer adjusting the documentation and defaults instead of the template.

GeertJohan:
@iljarotar Do your configurations perhaps define the variable twice, with and without underscore? Because it seems like the role asserts the underscore-version MUST be set? https://github.com/metal-stack/metal-roles/blob/master/partition/roles/mgmt-server/tasks/main.yaml#L8

GeertJohan:
It seems the linked assert always succeeds because

> Fable 5:
> Jinja's none test is just a Python value is None identity check, so an undefined variable passes is not none without error (the Undefined object is never actually touched, so it never raises — this holds even with StrictUndefined)

I've just pushed changes to fix the assert. Also implemented a backwards compatible move towards `router_id`, I thought it'd be nicer to have the proper spelling and cleanup elsewhere, but please let me know if you still prefer to stick with `routerid` then I'll move to that.

partition/roles/mgmt-server/tasks/main.yaml
```diff
@@ -4,11 +4,18 @@
...
+        (mgmt_server_routerid is defined and mgmt_server_routerid is not none)
+      - mgmt_server_spine_facing_interface is defined and mgmt_server_spine_facing_interface is not none
+      - mgmt_server_metal_ssh_privkey is defined and mgmt_server_metal_ssh_privkey is not none
+      - mgmt_server_metal_ssh_pubkey is defined and mgmt_server_metal_ssh_pubkey is not none
+
+- name: Warn about usage of deprecated variables
+  ansible.builtin.fail:
+    msg: "the variable mgmt_server_routerid was renamed to mgmt_server_router_id, please migrate and remove the old variable"
+  when: mgmt_server_routerid is defined
+  ignore_errors: true
```
Gerrit91: We have an own module for this, ansible-lint does not like ignore_errors.

https://github.com/metal-stack/metal-roles/blob/881728375de9a5bd12bb0c77c7e705eaba6c8221/partition/roles/promtail/tasks/main.yaml#L2-L10

GeertJohan: Ah great! Fixed.

Gerrit91:
To me these changes look good. If we add this breaking change to the release notes, it should be fairly easy to migrate (especially since you have this migration path in there). @iljarotar can you please take another look?

partition/roles/mgmt-server/tasks/main.yaml
```diff
@@ -4,11 +4,21 @@
...
+      - mgmt_server_metal_ssh_pubkey is defined and mgmt_server_metal_ssh_pubkey is not none
+
+- name: Deprecation warning for mgmt_server_routerid
+  run_once: true
+  print_message:
+    type: deprecate
+    msg: >-
+      The variable mgmt_server_routerid was renamed to mgmt_server_router_id.
+    action: Please define mgmt_server_router_id instead and remove the old variable.
+  when: mgmt_server_routerid is defined
```
iljarotar: Would it make sense to throw an error in this case?

iljarotar:
Are we leaving the deprecated `routerid` in there so folks have time to adapt and are we planning to remove it in a later release? Actually, if we opt for introducing this breaking change I wouldn't mind just breaking it without a transition period. If someone uses the role and only defines `routerid` then the fixed assert will tell them what's wrong.

GeertJohan:
Sounds good @iljarotar , I'll make the change.

GeertJohan:
Rebased and squashed.

iljarotar (approved):

iljarotar:
Please add a BREAKING_CHANGE notification in the PR description. This ensures that the breaking change is documented in the final release notes.

Looks like this:

\## Release Notes

\### Breaking Change

\```BREAKING_CHANGE
Description of the breaking change and what an operator needs to do about it.
This section is \*\*not\*\* intended for documentation of internal breaking changes.
Release notes are meant to be read by users and operators of metal-stack, not metal-stack developers.
\```

GeertJohan:
Added the breaking change line.

PR #632, opened 2026-06-18
ansible-lint: fix names.

Gerrit91:
## Description

Running ansible-lint autofix for [names](https://docs.ansible.com/projects/lint/rules/name/). Some names had to be manually fixed (providing a `name: ` for tasks that were missing a name or moving `{{ ... }}` to the end of the task name.

#### Used AI-Tools ✨

- None used for generation

PR #638, opened 2026-06-19
Fix task renaming regression: Apply renaming to handler names

iljarotar:
## Description

Fixes a regression of #632 where tasks were capitalized but not the handler names.


Gerrit91:
Sorry, this was my fault and is a regression of https://github.com/metal-stack/metal-roles/pull/632.

We need to check all handler names and also capitalize the first letter. I thought that ansible-lint --fix would take care of these kind of things but seems like it does not. :(

Gerrit91 (approved):

PR #641, opened 2026-06-23
Remove `docker-on-cumulus` role.

Gerrit91:
## Description

We do not support this anymore.

### Breaking Change

```BREAKING_CHANGE
The deprecated `docker-on-cumulus` role was removed.
```

#### Used AI-Tools ✨

- None used for generation


majst01:
But we still have switches in production with cumulus

Gerrit91:
We do not support it anymore and integration has been dropped from the mini-lab for long. When someone is still running this, they need to maintain the deployment on their own. It's not the responsibility anymore to maintain an untested role at this place.

majst01:
> We do not support it anymore and integration has been dropped from the mini-lab for long. When someone is still running this, they need to maintain the deployment on their own. It's not the responsibility anymore to maintain an untested role at this place.

I just fear that a production deployment is not possible anymore !

Gerrit91:
It's always possible to copy this role over if necessary by people who rely on it.

iljarotar (approved):

PR #643, opened 2026-06-23
ansible-lint: Fix risky-file-permissions

Gerrit91:
## Description

Now it's getting a bit more complex. I looked most of the permissions up on the partition servers and switches, so I think it should work fine.

Also there were some yaml errors left, which were somehow not reported even though only line-length should not be reported. This is confusing, but I now just provided a dedicated `.yamllint` file based on the default configuration for ansible-lint and increased the line-length there.

#### Used AI-Tools ✨

- None used for generation


partition/roles/promtail/tasks/main.yaml
```diff
@@ -32,6 +32,7 @@
   ansible.builtin.template:
     src: promtail.yaml.j2
     dest: "{{ promtail_config_host_dir }}/promtail.yaml"
+    mode: "0755"
```
simcod: Currently, we have `0644` here. But the role as the role is deprecated, it should be fine.

simcod (commented):
Looks good from my side. Maybe @iljarotar can have a look on the SONiC related parts.

partition/roles/promtail/tasks/main.yaml
Gerrit91: Perfect, thanks. I adjusted the value.

partition/roles/image-cache/tasks/coredns.yaml
```diff
@@ -3,6 +3,7 @@
   ansible.builtin.file:
     path: "{{ item }}"
     state: directory
+    mode: "0666"
```
majst01, line 6: Why a mixture of 666,644,640,755 ?

Gerrit91: Consistency because in the next task "0666" was used for the coredns config.

iljarotar:
sonic and metal-core stuff looks correct to me.

iljarotar (approved):

l0wl3vel:
@Gerrit91  you might want to revert this PR, as I am currently observing build failures in mini-lab due to 644 for directories preventing traversal for non-owners. Specifically in regards to the prometheus rules directory: https://github.com/metal-stack/metal-roles/blob/master/partition/roles/monitoring/prometheus/tasks/main.yaml#L185

Gerrit91:
Thanks for pointing it out. I will first attempt to forward fix and provide another PR for this.

PR #645, opened 2026-06-23
Deploy valkey with backup-restore-sidecar.

### LINKED ISSUE #634: Deploy valkey with backup-restore-sidecar
### Description

Currently we deploy it through helm-chart, but this does not create backups. We can utilize valkey support of our https://github.com/metal-stack/backup-restore-sidecar.

Gerrit91:
## Description

Closes #634.

### Required Actions

```ACTIONS_REQUIRED
Only for operators who deployed the metal-apiserver prematurely: The valkey deployment helm chart was removed and instead wrapped into deployment manifests using the backup-restore-sidecar project for backups. Therefore, it is necessary to uninstall the old valkey chart prior to the deployment using `helm delete valkey`. Existing tokens will not work anymore and need to be re-issued.
```

#### Used AI-Tools ✨

- None used for generation


control-plane/roles/valkey/defaults/main.yaml
```diff
@@ -1,17 +1,59 @@
...
 
-valkey_replicas: 3
-valkey_password: change-me-soon
+valkey_image_pull_policy: "{{ metal_control_plane_image_pull_policy }}"
 
-valkey_size: 10Gi
-valkey_storage_class: ""
+# TODO: move to release vector:
+valkey_image_name: ghcr.io/valkey-io/valkey
+valkey_image_tag: 9.1-alpine
```
majst01: i think this must be done in the next PR

majst01 (approved):

control-plane/roles/valkey/defaults/main.yaml
Gerrit91: Added to release vector, so I hope it works.^^

PR #647, opened 2026-06-24
feat: loki upgrade, persistence and retention

ma-hartma:
## Description

Currently our defaults store logs send to Loki in a directory in /tmp/. This is a tmpfs that is lost whenever the Container restarts. 

This PR aims to 
- [x] make log storage persistent
- [x] allow log retention
- [x] find a suitable default value for log retention
- [x] use the newest version of loki
- [x] monitoring, dashboards and alerts for loki
- [x] test in mini-lab
- [x] test in demo-rack 
- [ ] test if disabling persistence via `logging_loki_persistence_enabled` works
- [ ] test if logs are really gone after retention period

For now, we only add filesystem storage via helm-managed PVC.
Object-storage support is planned: https://github.com/metal-stack/metal-roles/issues/648 

We unfortunately cannot really offer a migration path for the log data, as the data is currentlylost on every container restart anyway - and changing anything leads to a container restart. 

Some used resources:
- https://archive.grafana.com/docs/loki/v2.8.x/installation/helm/reference/
- https://github.com/grafana/loki/issues/9207 
- https://archive.grafana.com/docs/loki/v2.8.x/operations/storage/retention/
- https://archive.grafana.com/docs/loki/v2.8.x/operations/storage/logs-deletion/

Closes #621 
Closes #613

#### Used AI-Tools ✨
- Claude Sonnet 4.6 for docs

References:

- https://github.com/metal-stack/mini-lab/pull/312 use in mini-lab
- https://github.com/metal-stack/releases/pull/311

PR #659, opened 2026-07-06
Configuration parameter for control-plane monitoring

simcod:
## Description

This pull request introduces more parameters for configuring the control-plane monitoring stack.

#### Used AI-Tools ✨

- None


control-plane/roles/monitoring/tasks/main.yaml
ma-hartma, line 99: I know it wasn't there before, but could you add a newline here?

control-plane/roles/monitoring/templates/prometheus-stack-values.yaml
```diff
@@ -39,10 +55,12 @@ prometheus:
...
           key: objstore.yml
           name: thanos-objstore-secret
+{% endif %}
 {% endif %}
   thanosService:
-    enabled: true
+    enabled: {{ monitoring_thanos_enabled }}
 
+{% if monitoring_grafana_enabled %}
 grafana:
```
ma-hartma, line 63: I think this must be grafana: enabled: false instead of leaving the part out, as of the default enabled: true. https://artifacthub.io/packages/helm/prometheus-community/kube-prometheus-stack/82.4.3?modal=values&path=grafana.enabled

ma-hartma (changes requested):
The idea is good and it makes sense, as different users have different requirements for remote writes.

control-plane/roles/monitoring/templates/prometheus-stack-values.yaml
```diff
@@ -310,12 +335,16 @@ kubeControllerManager:
 kubelet:
   enabled: {{ monitoring_prometheus_kubelet_enabled | lower }}
 
-{% if not monitoring_prometheus_kubelet_enabled %}
 defaultRules:
+  create: {{ monitoring_prometheus_default_rules_enabled }}
+{% if monitoring_prometheus_default_rules_enabled %}
```
ma-hartma, line 340: If create is conditional already, do we really need the second `if monitoring_prometheus_default_rules_enabled`?

control-plane/roles/monitoring/templates/prometheus-stack-values.yaml
```diff
@@ -39,11 +55,14 @@ prometheus:
...
           name: thanos-objstore-secret
+{% endif %}
 {% endif %}
   thanosService:
-    enabled: true
+    enabled: {{ monitoring_thanos_enabled }}
 
 grafana:
+  enabled: {{ monitoring_grafana_enabled }}
+{% if monitoring_grafana_enabled %}
```
ma-hartma, line 65: Same here: Extra condition necessary?

control-plane/roles/monitoring/templates/prometheus-stack-values.yaml
```diff
@@ -90,8 +111,11 @@ grafana:
 {% endif %}
   plugins:
     - camptocamp-prometheus-alertmanager-datasource
+{% endif %}
 
 alertmanager:
+  enabled: {{ monitoring_alertmanager_enabled }}
+{% if monitoring_alertmanager_enabled %}
```
ma-hartma, line 118: Same here? If there are defaults for everything are there, disabling should be enough as well.

control-plane/roles/monitoring/templates/prometheus-stack-values.yaml
simcod: With this condition the content will not be rendered, if `monitoring_grafana_enabled` is set to `false`. I thought it would make sense, to remove it in this case.

simcod: @Gerrit91 What do you think?

Gerrit91: No special opinion on this, from my side either way is fine. You need to come to an agreement. :)

ma-hartma: Then let's keep the extra conditionals, making the resulting config password-free e.g. for Grafana. I'll adapt in https://github.com/metal-stack/metal-roles/pull/647/changes as well ;)

ma-hartma (approved):

Issue #661, opened 2026-07-08
Cross-referenced ansible headscale variable


simcod:
### Description

In the default variables of the `metal-roles/control-plane/roles/metal` role there is a dependency to the `headscale_httproute_enabled` variable.
If the `headscale` role is not used, the deployment fails.

https://github.com/metal-stack/metal-roles/blob/45a942be7ec5ed6efe657f5e08027155c7b99abe/control-plane/roles/metal/defaults/main.yaml#L194-L199

PR #662, opened 2026-07-08
Fix cross-referenced headscale variable

simcod:
## Description

References: #661 

#### Used AI-Tools ✨

- None

### DIFF
--- a/control-plane/roles/metal/defaults/main.yaml
+++ b/control-plane/roles/metal/defaults/main.yaml
@@ -192,7 +192,7 @@ metal_api_headscale_tls: true
 metal_api_headscale_api_key: "{{ lookup('k8s', api_version='v1', namespace=metal_control_plane_namespace, kind='Secret', resource_name='headscale-api-key').get('data',
   {}).get('key') | b64decode if metal_api_headscale_enabled else '' }}"
 metal_api_headscale_control_plane_address: >-
-  {% if metal_api_httproute_enabled -%}
+  {% if headscale_httproute_enabled | default(false) -%}
   http{{ 's' if metal_api_headscale_tls }}://headscale.{{ metal_control_plane_gateway_dns }}
   {%- else -%}
   http{{ 's' if metal_api_headscale_tls }}://headscale.{{ metal_control_plane_ingress_dns }}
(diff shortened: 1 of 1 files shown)


control-plane/roles/metal/defaults/main.yaml
```diff
@@ -192,7 +192,7 @@ metal_api_headscale_tls: true
 metal_api_headscale_api_key: "{{ lookup('k8s', api_version='v1', namespace=metal_control_plane_namespace, kind='Secret', resource_name='headscale-api-key').get('data',
   {}).get('key') | b64decode if metal_api_headscale_enabled else '' }}"
 metal_api_headscale_control_plane_address: >-
-  {% if headscale_httproute_enabled -%}
+  {% if metal_api_httproute_enabled -%}
```
Gerrit91: This would assume that headscale is using the httproute already, but maybe this is not the case. How about:


```suggestion
  {% if headscale_httproute_enabled is not defined %}
  {{- None -}}
  {% elif headscale_httproute_enabled -%}
```

This would set `metal_api_headscale_control_plane_address: null`.

control-plane/roles/metal/defaults/main.yaml
```diff
@@ -192,7 +192,7 @@ metal_api_headscale_tls: true
...
   http{{ 's' if metal_api_headscale_tls }}://headscale.{{ metal_control_plane_gateway_dns }}
   {%- else -%}
   http{{ 's' if metal_api_headscale_tls }}://headscale.{{ metal_control_plane_ingress_dns }}
```
l0wl3vel: Unsure why the `metal_api_httproute_enabled` would have any impact on if headscale is exposed on gateaway or ingress.

```suggestion
  {% if headscale_httproute_enabled | default(false) -%}
  http{{ 's' if metal_api_headscale_tls }}://headscale.{{ metal_control_plane_gateway_dns }}
  {%- else -%}
  http{{ 's' if metal_api_headscale_tls }}://headscale.{{ metal_control_plane_ingress_dns }}
```

l0wl3vel (changes requested):
see comment

l0wl3vel (approved):
lgtm

Gerrit91 (approved):

PR #665, opened 2026-07-09
fix: add role prefix to event_exporter_enabled variable

ma-hartma:
## Description

Unfortunately I left out the `monitoring_` prefix for the `event_exporter_enabled` variable, meaning the default won't be false & the event exporter not removed.


Gerrit91:
Please add a breaking change release note.

ma-hartma:
Seems like it actually does work, and there was only a caching problem with the pipeline when I tried. 
image

PR #673, opened 2026-07-16
first version of automating user creation on ontap cluster

Honigeintopf:
In order to automate to control what users exist in the ontap netapp clusters we have to use ansible


majst01:
IMHO this should be a role which is located in a customer repository, e.g. metal-extension-roles

Honigeintopf:
Will be done in Customer Repo

PR #675, opened 2026-07-20
Deploy metal-metrics-exporter v2

majst01:
## Description

~Does create tenant yet because metal-v2-modules are not merged yet~

#### Used AI-Tools ✨

- none used for generation


control-plane/roles/monitoring/tasks/main.yaml
```diff
@@ -84,6 +84,42 @@
     helm_value_file_template: "thanos-values.yaml"
   when: monitoring_thanos_enabled
 
+- name: Create metal-apiserver token for metal-image-cache-sync
+  ansible.builtin.import_role:
+    name: metal-roles/common/roles/metal-deployment-token
```
Gerrit91, line 89: We usually expect this role to be executed from the playbook once.

vknabel: Especially as we already ensure this in the next task `Check metal-deployment-token has run`.

control-plane/roles/monitoring/templates/metrics-exporters/metal-metrics-exporter.yaml
```diff
@@ -1,13 +1,69 @@
...
+metadata:
+  name: metal-token-renewal
+rules:
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["get", "update", "patch"]
+- apiGroups: [""]
+  resources: ["secrets"]
+  resourceNames: ["metal-metrics-exporter-config"]
+  verbs: ["get", "update", "patch"]
```
vknabel, line 18: I think the second rule should be sufficient

PR #676, opened 2026-07-22
fix(alloy): allow setting docker network

ma-hartma:
## Description

It was expected, that docker bridges have Masquerading enabled and alloy containers can access the internet.
Unfortunately, a lot of switches do not support this - so this allows setting the network and eventually using the host network.

@simcod @Sven-Ric @iljarotar 

#### Used AI-Tools ✨

- none


simcod (approved):

Gerrit91 (approved):
Maybe a small hint in the docs why it has to be run in the host network would be great.

Gerrit91 (approved):

iljarotar (approved):

PR #687, opened 2026-07-30
Enable vpn in metal-apiserver

majst01:
## Description

Depends on: https://github.com/metal-stack/helm-charts/pull/173

#### Used AI-Tools ✨

- None used for generation

### DIFF
--- a/control-plane/roles/metal/defaults/main.yaml
+++ b/control-plane/roles/metal/defaults/main.yaml
@@ -84,6 +84,18 @@ metal_apiserver_tokens:
     tokenCreateRequest:
       expires: 23040s
       adminRole: ADMIN_ROLE_VIEWER
+metal_apiserver_headscale_enabled: false
+metal_apiserver_headscale_tls: true
+metal_apiserver_headscale_api_key: "{{ lookup('k8s', api_version='v1', namespace=metal_control_plane_namespace, kind='Secret', resource_name='headscale-api-key').get('data',
+  {}).get('key') | b64decode if metal_api_headscale_enabled else '' }}"
+metal_apiserver_headscale_control_plane_address: >-
+  {% if headscale_httproute_enabled | default(false) -%}
+  http{{ 's' if metal_api_headscale_tls }}://headscale.{{ metal_control_plane_gateway_dns }}
+  {%- else -%}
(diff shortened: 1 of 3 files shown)


control-plane/roles/metal/defaults/main.yaml
```diff
@@ -84,6 +84,17 @@ metal_apiserver_tokens:
     tokenCreateRequest:
       expires: 23040s
       adminRole: ADMIN_ROLE_VIEWER
+metal_apiserver_headscale_enabled: false
+metal_apiserver_headscale_tls: true
```
Gerrit91: Missing in documentation.

control-plane/roles/metal/defaults/main.yaml
```diff
@@ -84,6 +84,17 @@ metal_apiserver_tokens:
...
+metal_apiserver_headscale_api_key: "{{ lookup('k8s', api_version='v1', namespace=metal_control_plane_namespace, kind='Secret', resource_name='headscale-api-key').get('data',
+  {}).get('key') | b64decode if metal_api_headscale_enabled else '' }}"
+metal_apiserver_headscale_control_plane_address: >-
+  {% if headscale_httproute_enabled | default(false) -%}
+  http{{ 's' if metal_api_headscale_tls }}://headscale.{{ metal_control_plane_gateway_dns }}
+  {%- else -%}
+  http{{ 's' if metal_api_headscale_tls }}://headscale.{{ metal_control_plane_ingress_dns }}
+  {%- endif %}
+metal_apiserver_headscale_internal_api_address: "headscale:50443"
```
Gerrit91, line 97: Can just be hardcoded because how could this be changed to something different that would successfully work?

majst01: same as for metal-api before ?

control-plane/roles/metal/defaults/main.yaml
majst01: ok, fixed

control-plane/roles/metal/defaults/main.yaml
Gerrit91: Mh, maybe for unusual namespace choices... alright.

Gerrit91 (approved):

PR #688, opened 2026-07-30
mgmt-server: add optional static routes to the frr.conf template

GeertJohan:
The mgmt-server role owns and regenerates /etc/frr/frr.conf, so a deployment that needs a static route in FRR has no first-class place to put it. Adding it out-of-band does not work either: a vtysh command plus write memory after the role overwrites the role-rendered file with the running config, silently rolling back every role-driven FRR change on the next apply.

Add mgmt_server_frr_static_routes, a list of 'prefix nexthop' strings rendered as ip route entries. The default is empty and renders nothing, so existing deployments are byte-identical.

Example use case: a partition whose PXE VLAN is routed to the mgmt server over the fabric needs a return route whose recursive next-hop must be resolved by FRR rather than the kernel.

#### Used AI-Tools ✨

Claude Fable 5


iljarotar (approved):

iljarotar:
Need to fix Markdown formatting.

GeertJohan:
Rebased on lastest master

PR #701, opened 2026-07-31
Add configuration options for auditing and valkey

simcod:
## Description

This pull request adds missing Ansible variables to configure the s3 backend for the `auditing` and `valkey` databases.

Todo:

- [x] Document role variables for `valkey`

#### Used AI-Tools ✨

- None


control-plane/roles/valkey/defaults/main.yaml
```diff
@@ -24,6 +24,8 @@ valkey_backup_restore_sidecar_s3_secret_key:
 valkey_backup_restore_sidecar_s3_insecure_skip_verify:
 valkey_backup_restore_sidecar_s3_trusted_ca_cert:
 valkey_backup_restore_sidecar_object_max_keep: 480
+valkey_backup_restore_sidecar_object_days_max_keep:
```
Gerrit91: Probably we should add `_s3_` to this option as for other providers this flag does not have any effect? 

I guess it turned out that we should have prefixed all backup provider options because some providers do not implement them. Do not put too much effort into this though, it is not (yet) worth to write an expensive migration because of that. I would just already use the name `valkey_backup_restore_sidecar_s3_object_days_max_keep` for future reference.

simcod: Yes, you are right. This would be better. I changed the variable name for valkey and created an issue: https://github.com/metal-stack/metal-roles/issues/720

Gerrit91: Thanks. :)

PR #705, opened 2026-08-04
Validate rendered service configs before installing them

GeertJohan:
#### Used AI-Tools ✨

Claude Fable 5

### DIFF
--- a/partition/roles/frr/tasks/main.yaml
+++ b/partition/roles/frr/tasks/main.yaml
@@ -56,5 +56,6 @@
   ansible.builtin.copy:
     content: "{{ frr_conf }}"
     dest: /etc/frr/frr.conf
+    validate: "/usr/bin/vtysh --dryrun --inputfile %s"
     mode: "0640"
   notify: Reload frr
--- a/partition/roles/sonic/tasks/main.yaml
+++ b/partition/roles/sonic/tasks/main.yaml
@@ -58,6 +58,7 @@
   ansible.builtin.template:
     src: iptables.json.j2
     dest: /etc/sonic/iptables.json
+    validate: /usr/bin/python3 -m json.tool %s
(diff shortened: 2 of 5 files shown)


partition/roles/sonic/tasks/main.yaml
```diff
@@ -58,6 +58,7 @@
   ansible.builtin.template:
     src: iptables.json.j2
     dest: /etc/sonic/iptables.json
+    validate: /usr/bin/python3 -m json.tool %s
```
iljarotar: This file can be left out as the role is deprecated.

partition/roles/frr/tasks/main.yaml
```diff
@@ -56,5 +56,6 @@
   ansible.builtin.copy:
     content: "{{ frr_conf }}"
     dest: /etc/frr/frr.conf
+    validate: "/usr/bin/vtysh --dryrun --inputfile %s"
```
iljarotar, line 59: I haven't found any documentation about the --inputfile flag. Is that version-specific?

GeertJohan: Not version-specific. -f, --inputfile is one of vtysh's original getopt options, it is just missing from the man page.

It is in vtysh --help:
```
-f, --inputfile          Execute commands from specific file and exit
-C, --dryrun             Check configuration for validity and exit
```

partition/roles/sonic/tasks/main.yaml
GeertJohan: Removing..

partition/roles/frr/tasks/main.yaml
iljarotar: Amazing

iljarotar (approved):

PR #707, opened 2026-08-04
Add argument_specs to roles for typed variable validation

GeertJohan:
## Add role argument specs for typed variable validation

Ansible validates nothing about role variables by default: a typo'd optional variable is silently ignored, a string where a list is expected fails deep inside a template, and "mandatory" is only enforced where a hand-written assert happens to exist. This PR adds `meta/argument_specs.yaml` to every role (55 specs: partition, control-plane, common), so every variable is declared with a type, a description, and (where enforceable) `required: true`, validated by ansible before the first task runs. As a side effect, `ansible-doc` can now render per-role documentation from the specs.

### Design decisions

- **`required: true` is presence-based.** Empirically (ansible-core 2.21), a variable declared as `None` in defaults *satisfies* `required: true`, so requiredness is only truly enforced for variables with no defaults entry. None-defaulted-but-asserted variables are still marked required as documentation of intent; their runtime asserts remain the enforcement.
- **Release-vector variables are never `required`.** Image/chart variables resolved from `metal_stack_release.mapping` are set by `setup_yaml` at task time; *after* spec validation runs. Their specs note the resolution in the description instead. Asserts guarding them are kept.
- **Asserts were only removed where the spec fully replaces them** (6 pure presence-check blocks, two of which were literal duplicates). All richer asserts (value choices, XOR/conditional checks, placeholder checks, release-vector checks) are kept.
- **No `suboptions` on list/dict variables** unknown suboption keys would hard-fail existing inventories; structures stay loosely typed for now. **Secrets** (passwords, HMAC keys, private keys) carry `no_log: true`.
- Int/bool-typed variables that legitimately pass through as `None` are declared `raw` (type conversion rejects None, unlike str/list/dict).

### Verification

- All 55 specs load through ansible's `ArgumentSpecValidator`; each role's own defaults validate against its spec; every defaults key is covered.
- Running a role without its required variables fails with `missing required arguments: ...` before any task executes.
- [ ] TODO would be to run this version on actual hardware. I can use Fundament POC for all the roles we use there, but haven't done that yet.

### Changes that I think we should discuss

- Newly *enforced*. These have no default and no prior assert, but we probably should confirm deployments set them: `metal_core_cidr`; `metal_api_grpc_certs_{ca_cert,server_cert, server_key}`, `metal_masterdata_api_tls_*` and `headscale_noise_private_key`.
- `metal_api_grpc_certs_{ca_cert,server_cert,server_key}` are marked `required: true`. This codifies existing behavior rather than adding a new constraint: `metal-values.j2` renders these three values unconditionally (`| b64encode` fails on undefined), so every deployment must already define them, even with `metal_api_grpc_tls_enabled: false`, since the toggle is not used as a guard in the template. The spec just moves the failure from a mid-template error to a clear "missing required arguments" at role start. Probably `metal_api_grpc_tls_enabled` should be removed as it is currently unsable? But didn't want to scope-creep this (already huge) PR.
- Spec validation templates listed defaults at role start, so a few `lookup('k8s', ...)` defaults in the metal/monitoring roles evaluate slightly earlier than before (same role execution, earlier moment).
- Type declarations worth a second opinion: `mgmt_server_asn: int` (breaks dotted ASN notation), `dhcp_relay_server: raw` (string-or-list), sonic port speed/MTU ints.
- Found that `metal_core_hmac_key` appears unused and `mgmt_server_firewall_facing_interface` was documented as mandatory but no longer consumed (declared optional). Can we remove both?
- Also closed a gap: `metal_apiserver_image_name/tag` and `metal_metrics_exporter_image_name/tag` were rendered unguarded but missing from their roles' release-vector assert blocks. New checks added there, since argument_specs cannot enforce release-vector variables (see design decisions).

#### Used AI

Claude Fable 5

### DIFF
--- a/common/roles/metal-deployment-token/meta/argument_specs.yaml
+++ b/common/roles/metal-deployment-token/meta/argument_specs.yaml
@@ -1,0 +1,18 @@
+---
+argument_specs:
+  main:
+    short_description: Reads the metal-apiserver admin token from a secret and provides it as metal_deployment_admin_token
+    options:
+      metal_deployment_token_secret_name:
+        type: str
+        description: Name of the secret containing the metal-apiserver admin token.
+      metal_deployment_token_secret_namespace:
+        type: str
+        description: Namespace of the secret containing the metal-apiserver admin token.
+      metal_deployment_token_secret_key:
+        type: str
+        description: Key within the secret data that contains the token.
+      metal_deployment_admin_token:
+        type: str
+        no_log: true
+        description: The admin token to use for deployment. When unset, it is read from the referenced secret.
--- a/common/roles/metal-v2-client/meta/argument_specs.yaml
+++ b/common/roles/metal-v2-client/meta/argument_specs.yaml
@@ -1,0 +1,11 @@
+---
+argument_specs:
+  main:
+    short_description: Installs the metal-stack-api python client library
+    options:
+      metal_v2_client_install_latest_on_version_error:
+        type: bool
+        description: Install the latest available client as fallback when the requested version is not available.
+      metal_v2_client_install_from_git_repository:
+        type: bool
+        description: Install the client from the git repository instead of PyPI.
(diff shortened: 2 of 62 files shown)


Gerrit91:
Really cool! I did not even know this exists now. :scream: 

Not sure how to review this, I think I get tired already after three files. Do you think it would be possible to put this into smaller PRs?

PR #708, opened 2026-08-04
lint: fix no-changed-when and schema violations and enforce them

GeertJohan:
- add changed_when to all command/shell/raw tasks flagged by no-changed-when, using changed_when: false for read-only commands and changed_when: true for commands that always apply changes
- fix schema[meta] violations: drop the optional role_name field from galaxy_info (it is not used when roles are imported via their directory paths), add a missing description to the top-level meta/main.yml and replace invalid platform entries (SONiC is not a valid galaxy platform)
- add a missing meta/main.yml to the gnmic role so that the setup_yaml action from ansible-common can be resolved (fixes syntax-check[unknown-module])
- remove no-changed-when and schema from the ansible-lint warn_list so they are enforced from now on; role-name and var-naming stay warnings because renaming roles or variables would break downstream consumers
- enable forbid-duplicated-merge-keys in .yamllint (no violations)

#### Used AI-Tools ✨

Claude Fable 5

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -10,9 +10,8 @@ exclude_paths:
   # these are template results, so they do not need formatting:
   - partition/roles/sonic/test/data/**/metal.yaml
 
-# TODO: we need to gradually fix these!
+# role-name and var-naming are intentionally kept as warnings because renaming
+# roles or variables would break downstream consumers of these roles.
 warn_list:
   - role-name
-  - schema
   - var-naming
-  - no-changed-when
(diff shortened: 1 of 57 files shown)


vknabel:
Review missing from @metal-stack/metal-roles-maintainers-isolated-clusters and @metal-stack/metal-roles-maintainers-partition

vknabel (approved):

mwennrich (approved):

PR #709, opened 2026-08-04
Pin dependencies to commit hashes

GeertJohan:
We should probably discuss this..

As a security precaution we could consider pinning dependencies like this. Recent hacks (npm) show malicious code injected in dependency projects via the central repository. I guess ansible-galaxy can be hit by such an attack as well.

As a side effect of this no (accidentally) breaking changes by dependency projects are suddenly introduced and we can test every update. Now of course the downside of this is that we'll need to manually bump the version every time a new release is made, ideally after confirming the new release does not contain malicious changes.

#### Used AI-Tools ✨

Claude Fable 5

### DIFF
--- a/.github/workflows/oci-artifact.yaml
+++ b/.github/workflows/oci-artifact.yaml
@@ -17,10 +17,10 @@ env:
 
 jobs:
   draft:
-    uses: metal-stack/actions-common/.github/workflows/release-drafter.yaml@v1
+    uses: metal-stack/actions-common/.github/workflows/release-drafter.yaml@999a2bd0d754d7b71cd4d4359d20473e78862a4b # v1.0.6
 
   oci-ansible-role:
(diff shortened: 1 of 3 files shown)


Gerrit91:
I am not 100% convinced if we should really do it because it is a lot of effort to handle dependencies like this in so many repositories at least without AI? I'm open for this in places like [actions-common](https://github.com/metal-stack/actions-common), which is hopefully be consumed by all repositories at some point in time (spelling can already be consumed from there). When there's just a single place to maintain dependencies like this, I think it can be manageable.

More opinions would be nice.

vknabel:
@GeertJohan we decided to not pin everything to the commit hashes _for now_ as this would just be a solution for _this repo only_ and harder to maintain for now (especially across the whole organization). Instead we'd like to tackle the whole topic of automating dependency updates consistently across all repositories and ultimately being able to pin these dependencies.

The epic for this will be https://github.com/metal-stack/releases/issues/318. It's definitely on our roadmap and a general pain point.

PR #712, opened 2026-08-06
Store s3 request checksum in secret

simcod:
## Description

Follow-up of #633 
Gardener stores the `s3-request-checksum-calculation` configuration parameter in a `Secret` instead of a `ConfigMap`. We need to decide, if we want to refactor this and do it the same way.

TODO:

- [ ] Implement for rethinkdb as well

#### Used AI-Tools ✨

- None

### DIFF
--- a/control-plane/roles/postgres-backup-restore/templates/postgres.yaml
+++ b/control-plane/roles/postgres-backup-restore/templates/postgres.yaml
@@ -195,6 +195,13 @@ spec:
               key: insecureSkipVerify
               name: backup-restore-sidecar-backup-provider-config-{{ postgres_name }}
 {% endif %}
+{% if postgres_backup_restore_sidecar_s3_request_checksum_calculation %}
+        - name: BACKUP_RESTORE_SIDECAR_S3_REQUEST_CHECKSUM_CALCULATION
+          valueFrom:
+            secretKeyRef:
+              key: requestChecksumCalculation
+              name: backup-restore-sidecar-backup-provider-config-{{ postgres_name }}
+{% endif %}
 {% if postgres_backup_restore_sidecar_s3_trusted_ca_cert %}
         - name: BACKUP_RESTORE_SIDECAR_S3_TRUSTED_CA_CERT
           valueFrom:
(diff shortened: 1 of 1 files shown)


Gerrit91:
How do you use the backup-restore-sidecar in a Gardener context? This info would help me a lot because I was not aware of this and never considered it.

simcod:
> How do you use the backup-restore-sidecar in a Gardener context? This info would help me a lot because I was not aware of this and never considered it.

Sorry for the confusion. I do no use the `backup-restore-sidecar` component inside Gardener. However, while researching for all of the backup changes introduced recently, I also checked the `etcd-backup-restore-sidecar` implementation and just noticed,  that the `s3-request-checksum-calculation` configuration option is stored inside a `ConfigMap` instead of the `Secret`. I thought it could be considered as best-practice. But the change suggested in this pull request is optional.

control-plane/roles/postgres-backup-restore/templates/postgres.yaml
```diff
@@ -335,6 +339,9 @@ data:
 {% if postgres_backup_restore_sidecar_s3_trusted_ca_cert %}
   trustedCaCert: {{ postgres_backup_restore_sidecar_s3_trusted_ca_cert | b64encode }}
 {% endif %}
+{% if postgres_backup_restore_sidecar_s3_request_checksum_calculation %}
+    s3RequestChecksumCalculation: {{ postgres_backup_restore_sidecar_s3_request_checksum_calculation }}
```
Gerrit91, line 343: Wrong indent

Gerrit91:
Okay, I got it now. Not sure if it is really worth to change it as the field does not seem to reveal critical information. But if you think it's better, I am also open to accept it.

simcod:
> Okay, I got it now. Not sure if it is really worth to change it as the field does not seem to reveal critical information. But if you think it's better, I am also open to accept it.

Yes, just wanted to bring this up. But let's decide to keep it like it is. Especially, because this configuration option is not critical, so it should be fine to store it inside a `ConfigMap` as well.

